PR #1, opened 2016-08-02
Add the getWebfonts API

regiskuckaertz:
First and foremost, we need a way for our ads to use the guardian's webfonts. This is done via the `getWebfonts` service, that queries the page for webfonts and expects CSS `@font-face` declarations in return.


regiskuckaertz:
👍 

This is so beautiful I want to kiss my screen.

PR #2, opened 2016-08-02
Glabs Native Traffic Driver

regiskuckaertz:
Here's a rundown of what needs to happen for our first creative to be fully implemented:

- [x] create a native style in DFP
- [x] convert Scala template into DFP HTML template
- [x] import styles into DFP CSS template
- [x] get webfonts
- [ ] get GLabs-related URLs
- [x] deal with SVGs
- [x] deal with popup


regiskuckaertz:
Current status:
picture 2

Not bad, but:
- webfonts are not complete, this is because the webfonts in the markup are optimized for speedy delivery and only contain a single weight
- still need to figure out the SVGs

PR #20, opened 2016-09-19
Change PORT for running locally

JonNorman:
Some components may require that we simultaneously run frontend, which also runs on port 9000. Changing this repository is much easier than frontend :)

@regiskuckaertz, looking [here](https://github.com/guardian/commercial-templates/blob/jn-switch-preview-port/src/_shared/js/messages.js#L153) could we not remove the need for `DevMode` by just using `location.port`?


JonNorman:
Okay makes sense. Happy for me to merge this little guy in?

PR #24, opened 2016-09-22
Merchandising Travel component

regiskuckaertz:
Comes with a couple of perks:

- goodbye `DevMode`: I am now relying on the fact that the parent window will send its origin in the initial `getIframeId` call (will be available everywhere as soon as [this PR](https://github.com/guardian/frontend/pull/14419) goes through)
- timed messages: Promises are a low-level very limited API, that's why they're usually wrapped in libraries that add syntactic sugar for cancellation, timeouts, functional composition, etc. Such utility functions will be available in `js/impl/promises`. The first one, `timeout`, will reject if a promise does not resolve within the specified delay. I use it in `sendMessage`, first and foremost to prevent unavailable webfonts from leaving the creative blank; but it will be useful if `frontend` fails to answer one day or another. 300ms is a _very gracious_ amount, I'll work on that.

cc @JonNorman @annebyrne @Calum-Campbell @Ap0c @GHaberis


src/_shared/js/messages.js
```diff
@@ -16,11 +17,13 @@ export function getIframeId() {
...
                 json = JSON.parse(evt.data);
             } catch(_) { return; }
+
             const keys = Object.keys(json);
-            if( !(keys.length === 1 && keys[0] === 'id' )) return;
+            if( keys.length !== 2 || !keys.includes('id') || !keys.includes('host') ) return;
 
             self.removeEventListener('message', onMessage);
-            resolve(iframeId = json.id);
+            ({ id: iframeId, host: parentOrigin } = json);
```
JonNorman: Niiiiice

src/manual-gimbap/index.js
```diff
@@ -1,60 +1,50 @@
...
-
-	}
-
-}).catch(err => {
-	console.log(err);
-});
+getIframeId()
+.then(() => Promise.all([getWebfonts(), formatGimbap()]))
+.then(resizeIframeHeight);
+
```
JonNorman: Not worth catching/logging any errors? I suppose they will be shown in the console if they reject anyway....

src/travel/index.js
```diff
@@ -0,0 +1,42 @@
...
+.then(() => fetch(host + config.travelUrl))
+.then(response => response.json())
+.then(offers => offers.map(createAdvert))
+.then(html => write(() => container.innerHTML = html.join('')))
+.then(resizeIframeHeight);
+
+
+/* Outputs the HTML for a travel advert */
+function createAdvert(offer, index) {
+    return `<a class="blink advert advert--travel advert--prominent-${ index === 0 ? '[%IsProminent%]' : 'false'} ${ index > 1 ? 'hide-until-tablet' : '' }" href="%%CLICK_URL_UNESC%%${offer.offerUrl}" data-link-name="">
```
JonNorman: Ah I didn't know about JS HTML literals :( This will simplify mine...

src/travel/index.js
```diff
@@ -0,0 +1,42 @@
+import config from '../_shared/js/config';
+import { write } from '../_shared/js/dom';
+import { getIframeId, getWebfonts, resizeIframeHeight } from '../_shared/js/messages';
+
+let host = '[%DevMode%]' === 'false' ? 'https://www.theguardian.com' : 'http://localhost:9000'
```
JonNorman: Can we not now get rid of `DevMode` by getting the host from `messages.js`?

src/_shared/js/messages.js
regiskuckaertz: yeah destructuring is a godsend for you fingers 😆  I started learning standard ML and pattern matching is soooo easy to read.

src/travel/index.js
regiskuckaertz: oh, very good catch, I missed this one.

src/travel/index.js
regiskuckaertz: yeah my bad, I should've paired with you for the js 😞

src/manual-gimbap/index.js
regiskuckaertz: Not sure what the best course of action is here, happy to discuss. Indeed I thought the console might be a good last resort debugger, at least at the beginning.

Issue #21, opened 2016-09-20
Ignore Build Directory


Ap0c:
Dropping the build dir was discussed in #15. Is it worth adding to the `.gitignore`?

JonNorman:
I was just about to add this - I think we should ignore it. I'll raise a PR.

JonNorman:
https://github.com/guardian/commercial-templates/pull/25

PR #25, opened 2016-09-22
Removes contents of build directory from repository

JonNorman:
image


Ap0c:
Does the build process generate `preview/public/index.html`?

JonNorman:
@Ap0c: good point - I (mis)thought that both were auto generated. I'll add that back in.

Ap0c:
Awesome, fixes #21 too!

JonNorman:
Smashing.

PR #28, opened 2016-09-22
Provide a method to clear the build directory

JonNorman:
If the `build` task fails, then it fails silently. Instead of fixing that, I just wrote a `clean` task that can be run before running `build`, thus making it obvious when the build fails - i.e. an empty target folder - rather than having the old files squatting there and deceiving you into thinking it has correctly built.

### DIFF
--- a/clean.js
+++ b/clean.js
@@ -1,0 +1,20 @@
+const rimraf = require('rimraf');
+const Promise = require('bluebird');
+const fs = Promise.promisifyAll(require('fs'));
+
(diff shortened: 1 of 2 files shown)


JonNorman:
Now when I'm working on my templates I just run:
`npm run clean; npm run build; npm run build-preview`

and run `npm serve-preview` in another tab. Might be worth making a nodemon task to notice changes in the `src` directory. But then again, maybe not :)

PR #29, opened 2016-09-23
Native Books component

regiskuckaertz:
With a few changes:

- `getIframeId` will now resolve with the `id` and `host` values returned by the parent window. This is to allow querying the correct JSON endpoint ( @JonNorman you will probably have a conflict, sorry I had to do it 😊 )
- I forgot to take `NumberofCards` into account in the travel component
- I also forgot to take `IDs` into account in the travel component


regiskuckaertz:
Sure do! And perhaps raise a PR with your `portify` in the shared folder?

PR #36, opened 2016-09-26
Soulmates Native

JonNorman:
[Here it is.](https://www.youtube.com/watch?v=z5dpYDTEMRU)

**Old**
image

**Native**
image

This is the first template I've done so I'm fully prepared for a wave of comments and changes.

We could add in a parameter for the number of soulmates cards to render (we take 6 by default, but since we render each card as it comes, we could vary this and provide control of this in DFP. This functionality isn't currently present but it would be trivial to implement.

### DIFF
--- a/src/_shared/scss/_advert.scss
+++ b/src/_shared/scss/_advert.scss
@@ -19,11 +19,6 @@
     }
 }
 
-a.advert {
-  text-decoration: none;
-  cursor: pointer;
-}
-
 .advert--prominent-true, .advert--landscape {
     .has-no-flex & {
         overflow: hidden;
--- a/src/_shared/scss/_advert.scss
+++ b/src/_shared/scss/_advert.scss
@@ -19,6 +19,11 @@
     }
 }
 
+a.advert {
+  text-decoration: none;
+  cursor: pointer;
+}
+
 .advert--prominent-true, .advert--landscape {
     .has-no-flex & {
         overflow: hidden;
@@ -109,6 +114,7 @@
 
     &.u-text-hyphenate {
         max-width: 100%;
+        word-wrap: break-word;
     }
 
     .advert-blended__body & {
--- a/src/soulmates/index.js
+++ b/src/soulmates/index.js
@@ -1,0 +1,29 @@
+import { getIframeId, getWebfonts, resizeIframeHeight } from '../_shared/js/messages';
+import { write } from '../_shared/js/dom';
+import { portify } from '../_shared/js/dev';
+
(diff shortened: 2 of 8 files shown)


JonNorman:
Do we want to standardise our native templates API by making all of the requests send a `t` parameter? Currently this one just prepends SubGroup to the end of the URL...

src/_shared/scss/_advert.scss
```diff
@@ -19,6 +19,11 @@
     }
 }
 
+a.advert {
```
regiskuckaertz: Sorry, you can add the `blink` class to the `advert` and then `blink__anchor` to the `advert__title`

src/_shared/scss/_advert.scss
```diff
@@ -109,6 +114,7 @@
 
     &.u-text-hyphenate {
         max-width: 100%;
+        word-wrap: break-word;
```
regiskuckaertz: Yeah we should probably copy the `u-text-hyphenate` class from frontend in _core.scss

src/soulmates/index.js
```diff
@@ -0,0 +1,29 @@
...
+import { write } from '../_shared/js/dom';
+import { portify } from '../_shared/js/dev';
+
+const cardContainer = document.getElementsByClassName("adverts__row")[0];
+
+getIframeId()
+  .then(({host}) => fetch(`${portify(host)}${config.soulmatesUrl}`))
+  .then(response => response.json())
+  .then(soulmates => soulmates.map(createSoulmateCard))
+  .then(cards => addSoulmatesCards(cards))
```
regiskuckaertz: You can merge this line with the next one using `Promise.all`, this will go a little bit faster.

```
.then(cards => Promise.all([addSoulmatesCards(cards), getWebfonts()])
```

src/soulmates/index.scss
```diff
@@ -0,0 +1,4 @@
+@import '_core';
+@import '_adverts';
+@import '_adverts-soulmates';
+@import '_advert';
```
regiskuckaertz: I'm surprised you haven't seen more side effects. What happens if you put `_adverts-soulmates` last?

JonNorman: I can't see any changes from when it was the penultimate stylesheet getting imported, but I've moved it to the bottom.

JonNorman:
@regiskuckaertz happy for this to go out with the latest changes?

regiskuckaertz:
re the JSON URL, your call.

PR #43, opened 2016-09-29
CAPI Merchandising Single - Paid For Template

annebyrne:
This is the native template for the CAPI single paid-for template. I have split out paid-for and supported into two separate templates, for clarity this is only the paid-for template.

Now the whole card can be populated from 1 keyword or custom URL, with the exception of the component title. This incorporates recent work at pushing brand logos and data to CAPI.

cc @regiskuckaertz @JonNorman


src/_shared/scss/_core.scss
```diff
@@ -227,6 +234,10 @@ $font-scale: (
     font-weight: normal;
 }
 
+@mixin f-headlineSans {
```
regiskuckaertz: This can go, useless.

src/_shared/scss/_core.scss
```diff
@@ -10,6 +10,13 @@ $gs-column-width: 60px;
 $gs-row-height: 36px;
 $gs-max-columns: 16;
 
+// For popups
+$headerHeight: 36px;
```
regiskuckaertz: Why do you need these vars here?

src/_shared/scss/_core.scss
regiskuckaertz: Ok, I'll take care of this. All fonts mixins should go, they're boring.

src/_shared/scss/_core.scss
```diff
@@ -240,10 +251,47 @@ $font-scale: (
     line-height: convert-to-px($line-height);
 }
 
+// Hide content visually, still readable by screen readers
+@mixin u-h {
```
regiskuckaertz: You don't need this. If you do, please just add the class `u-h` to the element you want to hide.

src/_shared/scss/_popup.scss
```diff
@@ -0,0 +1,294 @@
+/* ==========================================================================
```
regiskuckaertz: This whole file is useless.

src/capi-single-paid/index.js
```diff
@@ -0,0 +1,73 @@
...
+let customUrl = '[%CustomUrl%]';
+
+if (customUrl !== '') {
+  params.append('t', customUrl);
+} else {
+  params.append('k', keywords);
+}
+
+const GLABS_EDITION = {
+	default: 'https://theguardian.com/guardian-labs',
```
regiskuckaertz: You can strip the `https://theguardian.com` part since I added it to the `BASE` element. All URIs that are not fully-qualified will open on the guardian.

annebyrne: I've removed this as the edition check will be better placed on the server side.

src/_shared/scss/_core.scss
annebyrne: 💥  AND....they're gone https://github.com/guardian/commercial-templates/pull/43/commits/a1b03e865731cf0a8b9aafcbc7803f88c47cdfe7

annebyrne:
@regiskuckaertz see https://github.com/guardian/commercial-templates/pull/43/commits/a1b03e865731cf0a8b9aafcbc7803f88c47cdfe7 💥  unwanted styling, begone!

regiskuckaertz:
tchoo tchoo 🚅

PR #44, opened 2016-09-29
ASK AND YOU SHALL RECEIVE

JonNorman:
(the ability to explicitly provide a minification directive to `build.js`)


JonNorman:
Whoop whoop!

PR #46, opened 2016-09-30
Remove font mixins

regiskuckaertz:


JonNorman:
LGTM :+1: (Read that as: this doesn't directly affect anything I've worked on and thus I'm fine with it)

PR #49, opened 2016-10-01
Blended component

regiskuckaertz:
Lo and behold, here is the blended component.

@JonNorman I had to change the way you deal with the soulmates URL, because config ends up being inserted into every component and the `SubFeed` variable only exists in the soulmates format, DFP complains about an inexistent variable in other components. Actually, I'm thinking this whole config object was a bad idea.


JonNorman:
Ah I think part of that was covered here? https://github.com/guardian/commercial-templates/pull/45

src/_shared/js/utils.js
```diff
@@ -0,0 +1,8 @@
+
```
JonNorman: Shall we move the `dev.js` methods here?

JonNorman:
Looks good to me - but could you take a look at my PR #45 before merging? Happy to go with either method of changing the soulmates endpoint (I've got a frontend PR open for this too https://github.com/guardian/frontend/pull/14531)

regiskuckaertz:
@JonNorman by Jove! Yours is better, please merge and I'll remove my changes.

src/_shared/js/utils.js
regiskuckaertz: Oh, I forgot about this one. Yeah slowly but surely the `js` and `scss` folders are getting unwieldy, let's think about this in a future comprehensive PR.

JonNorman:
Okay great! Merged my changes :)

PR #54, opened 2016-10-03
Srcset Bug Fix

annebyrne:
A combination of typos ( 😨  come on now Anne!), and poor refactoring meant that a bug was introduced where the srcset images weren't correctly populated and so that images would not render.

The secondary bug was in missing the write() import, which meant that whilst the templates worked in preview testing they haven't been rendering as a creative on site with a test line item.

cc @regiskuckaertz @Ap0c


src/_shared/js/images.js
```diff
@@ -1,33 +1,34 @@
...
 export function insertBetweenComments(sources) {
 
 	let pictures = Array.from(document.getElementsByTagName('picture'));
 
   return write(() => {
-		pictures.forEach((picture, index) => picture.insertBefore(sources[index], picture.firstChild));
+		pictures.forEach((picture, index) => {
+      picture.insertBefore(sources[index][0], picture.firstChild);
+      picture.insertBefore(sources[index][1], picture.firstChild);
+    });
```
Ap0c: I don't think this works... There's only one `picture` tag with six sources, not three `picture` tags with two sources each. As a result this only inserts the first two `source` elements into the page and skips out the last four.

src/capi-single-supported/index.js
```diff
@@ -62,7 +62,7 @@ function populateCard(responseJson) {
           </picture>
```
Ap0c: You need to have the IE9 comments already [present](https://github.com/guardian/commercial-templates/blob/capi-multiple/src/capi-multiple-paidfor/index.html#L44) in the page in order be inserting between them with the `insertBetweenComments` function.

annebyrne:
@regiskuckaertz @Ap0c 🚢   ?

src/_shared/js/images.js
```diff
@@ -18,18 +18,25 @@ export function addSourceset(responseJson) {
...
-	let pictures = Array.from(document.getElementsByTagName('picture'));
+    let picture = document.getElementsByTagName('picture')[0];
 
-  return write(() => {
-		pictures.forEach((picture, index) => {
-      picture.insertBefore(sources[index][0], picture.firstChild);
-      picture.insertBefore(sources[index][1], picture.firstChild);
+    return write(() => {
+      sources.forEach((source, index) => {
+        picture.insertBefore(source[0], picture.firstChild);
```
regiskuckaertz: Let's see what this gives us, but we might want to use `appendChild` instead of `insertBefore` here.

src/_shared/js/images.js
```diff
@@ -18,18 +18,25 @@ export function addSourceset(responseJson) {
...
+      });
     });
-	});
+  }
+
+export function buildImages(sources) {
+
+  let imageTag = document.getElementsByTagName('img')[0];
+
+  (imageTag.srcset === '')? insertSrcset(sources) : imageTag;
```
regiskuckaertz: Better: `'srcset' in imageTag ? ...`

src/_shared/js/messages.js
```diff
@@ -19,7 +19,7 @@ export function getIframeId() {
             } catch(_) { return; }
 
             const keys = Object.keys(json);
-            if( keys.length !== 3 || !keys.includes('id') || !keys.includes('host') ) return;
```
regiskuckaertz: Feels like this does not belong to this PR

src/_shared/scss/_adverts.scss
```diff
@@ -324,13 +324,8 @@
 
 .adverts--legacy-inline {
     padding: 0;
-    width: $gs-gutter * 6.5;
```
regiskuckaertz: Ibid.

src/manual-inline/index.html
```diff
@@ -0,0 +1,19 @@
+<aside class="adverts adverts--manual adverts--legacy-inline adverts--[%Tone%] adverts--tone-[%Tone%]" data-link-name="creative | ad inline manual | [%OmnitureId%]">
```
regiskuckaertz: Ibid.

src/manual-inline/index.js
```diff
@@ -0,0 +1,5 @@
+import { getIframeId, resizeIframeHeight,getWebfonts } from '../_shared/js/messages.js';
```
regiskuckaertz: Ibid.

src/_shared/js/images.js
Ap0c: We're going to refactor a lot of this together with the multiple in a separate PR. Currently just getting our versions merged so we're not working in parallel.

PR #57, opened 2016-10-04
Call getWebfonts explicitely without args

regiskuckaertz:
Otherwise it is receiving whatever `getIframeId` returns.


annebyrne:
LGTM 👍

PR #51, opened 2016-10-03
Capi Multiple

Ap0c:
# In This PR:

- cAPI Multiple Paidfor
- cAPI Multiple Supported
- Some supporting SASS files migrated from frontend.
- Some supporting SVGs migrated from frontend.
- A shared `js` file containing most of the code for the two templates.

# Pending:

- Tested in PROD behind ad cookies, but will wait for @regiskuckaertz 's resize listener PR before merging, as there's still minor cropping on window resize.

@annebyrne @JonNorman @Calum-Campbell

PR #58, opened 2016-10-04
Refactor Capis

Ap0c:
# In This PR:

Refactored a lot of common code related to image insertion into the module `capi-images.js`, shared between all capi templates.

# Changes:

- Created new shared module for capi images.
- Pulled out image insertion code from capi single and multiple templates; replaced with calls to the new image module.
- Renamed the `adverts__badge` class to `badge`.

# Comment:

This should cover your comments on duplicated code in both #51 and #54 @regiskuckaertz.
@annebyrne


src/capi-single-paid/index.js
```diff
@@ -25,8 +26,8 @@ enableToggles();
 getIframeId()
 .then(() => fetch(`https://api.nextgen.guardianapps.co.uk/commercial/api/capi-single.json?${params}`))
 .then(response => response.json())
-.then(capiData => [addSourceset(capiData.articleImage.sources), populateCard(capiData)])
-.then(([sources, html]) => Promise.all([getWebfonts(['GuardianTextSansWeb', 'GuardianSansWeb']), write(() => container.innerHTML = html)]).then(() => buildImages(sources)))
+.then(capiData => [capiData.articleImage, populateCard(capiData)])
+.then(([imageInfo, html]) => Promise.all([getWebfonts(['GuardianTextSansWeb', 'GuardianSansWeb']), write(() => container.innerHTML = html)]).then(() => addImage(imageInfo)))
```
annebyrne: Webfonts doesn't take any args anymore I believe following this PR: https://github.com/guardian/commercial-templates/pull/57

src/_shared/js/capi-images.js
```diff
@@ -0,0 +1,65 @@
+// Creates a set of source elements for an image.
+function buildSources (sourceData) {
+
+	let sourcesFragment = document.createDocumentFragment();
+
+	return sourceData.reduce((sources, source) => {
```
annebyrne: Would it make sense to use forEach here instead of reduce because of side-effects? @regiskuckaertz

src/capi-single-supported/index.js
```diff
@@ -19,7 +19,7 @@ getIframeId()
 .then(() => fetch(`https://api.nextgen.guardianapps.co.uk/commercial/api/capi-single.json?${params}`))
 .then(response => response.json())
 .then(capiData => [capiData.articleImage, populateCard(capiData)])
-.then(([imageInfo, html]) => Promise.all([getWebfonts(['GuardianTextSansWeb', 'GuardianSansWeb']), write(() => container.innerHTML = html)]).then(() => addImage(imageInfo)))
+.then(([imageInfo, html]) => Promise.all([getWebfonts(), write(() => container.innerHTML = html)]).then(() => addImage(imageInfo)))
```
annebyrne: This still won't work till this PR: https://github.com/guardian/commercial-templates/pull/59 is merged, but LGTM

src/capi-single-paid/index.js
Ap0c: Didn't know they ever did 😮, I never gave them any in the multiple... anyway, removed: 6bf423e.

src/_shared/js/images.js
```diff
@@ -1,44 +1,3 @@
...
-  }
-
-export function buildImages(sources) {
-
-  let imageTag = document.getElementsByTagName('img')[0];
-
-  "srcset" in imageTag ? insertSrcset(sources) : imageTag;
-}
-
 export function checkIcon(responseJson) {
```
Ap0c: b6db588

Ap0c:
Decided to drop the `adverts__badge` changes into this PR too since they're pretty minor and don't appear to affect the look of anything 88fe7af @regiskuckaertz @annebyrne.

src/_shared/js/capi-images.js
regiskuckaertz: Yes, there's no real reason to use a reducer with a DOM object. If you want to avoid access to the closure, you can pass `sourcesFragment` as the context in `forEach`, like that `sourceData.forEach(fn, sourcesFragment)` and inside `fn` you must reference `sourcesFragment` as `this`. This is by the way a performance tip that is used and abused in bluebird.

src/_shared/js/capi-images.js
```diff
@@ -0,0 +1,80 @@
...
+
+}
+
+// Creates a picture element with responsive sources, with fallback for IE.
+function createPicture (imageInfo, imageElem) {
+
+	// Supports responsive images.
+	if ('srcset' in imageElem) {
+
+		let picture = document.createElement('picture');
```
regiskuckaertz: Just thought about it, but in [this PR](https://github.com/guardian/commercial-templates/pull/60) I have created a HOF that abstracts the creation of DOM elements. Might be useful.

src/capi-single-paid/index.js
```diff
@@ -25,8 +25,8 @@ enableToggles();
 getIframeId()
 .then(() => fetch(`https://api.nextgen.guardianapps.co.uk/commercial/api/capi-single.json?${params}`))
 .then(response => response.json())
-.then(capiData => [addSourceset(capiData.articleImage.sources), populateCard(capiData)])
-.then(([sources, html]) => Promise.all([getWebfonts(['GuardianTextSansWeb', 'GuardianSansWeb']), write(() => container.innerHTML = html)]).then(() => buildImages(sources)))
+.then(capiData => [capiData.articleImage, populateCard(capiData)])
+.then(([imageInfo, html]) => Promise.all([getWebfonts(), write(() => container.innerHTML = html)]).then(() => addImage(imageInfo)))
```
regiskuckaertz: You could even have the `then` after the `write` rather than after the whole promise array.

src/_shared/js/capi-images.js
Ap0c: 0d66123

PR #59, opened 2016-10-04
Small CSS tweaks to CAPI Single templates

annebyrne:
- WebFonts were being retrieved but not rendered on CAPI supported. This was because the page had to wait for the HTML to be rendered
- The button on Supported was not the correct style - added background and border color
- The popup tag on CAPI paid was incorrect, adding correct tag and specific styling fixed this.

cc @regiskuckaertz

### DIFF
--- a/src/capi-single-supported/index.js
+++ b/src/capi-single-supported/index.js
@@ -19,7 +19,7 @@ getIframeId()
 .then(() => fetch(`https://api.nextgen.guardianapps.co.uk/commercial/api/capi-single.json?${params}`))
 .then(response => response.json())
 .then(capiData => [addSourceset(capiData.articleImage.sources), populateCard(capiData)])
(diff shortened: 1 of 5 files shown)


src/capi-single-supported/index.js
```diff
@@ -19,7 +19,7 @@ getIframeId()
 .then(() => fetch(`https://api.nextgen.guardianapps.co.uk/commercial/api/capi-single.json?${params}`))
 .then(response => response.json())
 .then(capiData => [addSourceset(capiData.articleImage.sources), populateCard(capiData)])
-.then(([sources, html]) => Promise.all([getWebfonts(['GuardianTextSansWeb', 'GuardianSansWeb']), write(() => container.innerHTML = html)]).then(() => buildImages(sources)))
+.then(([sources, html]) => Promise.all(write(() => container.innerHTML = html)).then(() => buildImages(sources)).then(getWebfonts()))
```
regiskuckaertz: Why the change of sequence? Besides, `Promise.all` takes an array of promises as argument.

PR #60, opened 2016-10-05
Fabric static with improvements for fixed/parallax cases

regiskuckaertz:
This prepares the Fabric template to handle all three cases of scrolling effects: none, fixed and parallax.

It also introduces several new/updated APIs:

- `write` and `read` now accept additional arguments, that will be passed to the callback. So, for example in `write(fn, arg1, arg2, arg3)`, `fn` will be called passing the three arguments.
- `write` will resolve to its return value. This to allow creating elements and returning references to them. The keen eye will notice that `write` and `read` are now exactly the same. The next update will introduce scheduling techniques that prioritises reads over writes.
- creating DOM elements has been abstracted, the idea is to provide functions of the form `tagName(props: Object?, attrs: Object?)`.


src/_shared/js/dom.js
```diff
@@ -13,15 +13,34 @@ if( ! window.requestAnimationFrame ) {
...
-        resolve(fn());
+        resolve(fn(...args));
     }));
 }
+
+function createElement(tagName) {
+    return function(props = {}, attrs = {}) {
+        let elem = document.createElement(tagName);
+        Object.assign(elem, props);
+        Object.keys(attrs).forEach(attr => img.setAttribute(attr, attrs[attr]));
```
Ap0c: Is this meant to be `elem.setAttribute`?

regiskuckaertz: Oops, good catch thanks 😅

src/_shared/js/dom.js
```diff
@@ -13,15 +13,34 @@ if( ! window.requestAnimationFrame ) {
...
+
+let img = createElement('img');
+let div = createElement('div');
+
+export {
+    closest,
+    write,
+    read,
+    img,
+    div
```
Ap0c: Just curious, why export these two elements? Is it because we're using them so often?

regiskuckaertz: I just thought about these two first, but the idea would be eventually to have all the HTML elements.

regiskuckaertz:
Thanks!

regiskuckaertz:
@Ap0c Indeed these changes are purely stylistic, there is no fundamental improvement. In your PR for example, you could write:

```
function addImage (imageInfo) {
  let imageContainer = document.querySelector('.advert__image-container');
  insertImage(imageContainer, imageInfo, '[%ArticleImage%]');
}
```

and then in your promise chain, just `.then(() => write(addImage, imageInfo)`

PR #62, opened 2016-10-05
Fix Capi Prod Errors

Ap0c:
# In This PR

Minor fixes noticed in PROD testing:

- JS selector was using a class that got renamed, fixed this (didn't show up locally for some reason).
- Updated text colouring to match pre-existing capi ads (think this was previously inheriting from `body`, no longer possible in iframe)

@regiskuckaertz @annebyrne @JonNorman @Calum-Campbell


Ap0c:
Good call, turns out it was in our palette.

PR #63, opened 2016-10-06
Fix for Chrome bug on CAPI Supported Single Template

annebyrne:
**Bug**
Chrome has a bug whereby requestAnimationFrame does not work with empty elements. 

Previously the CAPI Single Supported template had a single aside tag in its HTML, and the template was rendered by the javascript for simplicity due to header and body structure.This meant that the element had no height associated with it. 

**FIx**
The population of CAPI data in the header and the body of the component has been separated into 2 different functions and the HTML includes more content prior to JS rendering.

There have been a few small changes to shared JS files.

**Screenshots**

Firefox:

screen shot 2016-10-06 at 18 40 37

Chrome:

screen shot 2016-10-06 at 18 41 31

cc @guardian/commercial-dev @regiskuckaertz

### DIFF
--- a/src/_shared/js/messages.js
+++ b/src/_shared/js/messages.js
@@ -1,4 +1,4 @@
-import { read, write } from './dom.js';
+import { read, write } from './dom';
 import { timeout } from './impl/promises';
 
 const rootElement = document.documentElement;
@@ -81,7 +81,7 @@ function getWebfonts(fontFamilies) {
 }
 
 function resizeIframeHeight() {
-    return Promise.all([isDocumentLoaded()].concat(areImagesLoaded()))
+    return Promise.all(areImagesLoaded().concat(isDocumentLoaded()))
     .then(() => read(() => document.body.getBoundingClientRect().height))
     .then(function(height) {
         return sendMessage('resize', { height });
@@ -122,7 +122,7 @@ function sendMessage(type, value) {
                 reject(error);
             }
         });
-
+        console.log("sending message", id, iframeId, type, value);
         post(id, iframeId, type, value);
     }), 300);
 }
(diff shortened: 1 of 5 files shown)


src/capi-single-supported/index.js
```diff
@@ -14,12 +14,11 @@ if (customUrl !== '') {
...
 };
 
-enableToggles();
 getIframeId()
 .then(() => fetch(`https://api.nextgen.guardianapps.co.uk/commercial/api/capi-single.json?${params}`))
 .then(response => response.json())
-.then(capiData => [capiData.articleImage, populateCard(capiData)])
-.then(([imageInfo, html]) => Promise.all([getWebfonts(), write(() => container.innerHTML = html)]).then(() => addImage(imageInfo)))
+.then(capiData => [populateLogo(capiData), populateCard(capiData)])
+.then(([logo, body, imageInfo]) => Promise.all([getWebfonts(), write(() => header.innerHTML = logo, container.innerHTML = body).then(() => addImage(imageInfo))]))
```
regiskuckaertz: how come the array argument has three elements in it, if the previous callback returns an array of only two?

Ap0c: The third argument, `imageInfo`, is not being passed through here.

src/_shared/js/messages.js
```diff
@@ -122,7 +122,7 @@ function sendMessage(type, value) {
                 reject(error);
             }
         });
-
+        console.log("sending message", id, iframeId, type, value);
```
Ap0c: Minor thing, but is this intended or left over from debugging?

annebyrne:
@guardian/commercial-dev @regiskuckaertz Changes made to reflect typo.

PR #64, opened 2016-10-07
Handle Fabric fixed and parallax effects

regiskuckaertz:
Also fixes a bug where the callback listener was undefined. Related to [this PR](https://github.com/guardian/frontend/pull/14613).

@Ap0c @JonNorman @Calum-Campbell @annebyrne so you can update your creatives should you need it, when the viewport gets resized.


Ap0c:
Nice, the viewport stuff is going to be really useful for the capi multiple 👍

PR #65, opened 2016-10-07
Open links in a new tab

regiskuckaertz:
This is the standard behaviour for all our ads, plus it will allow us to track clicks more consistently.


regiskuckaertz:
(I'm thankful @JonNorman will write a command line tool to batch-update all native styles with the latest build 😉 )

JonNorman:
nothing

PR #67, opened 2016-10-10
Gimbap simple native

Calum-Campbell:
- Gimbap simple component added to native templates
- Holiday component tones added
- Money tones added
- .has-flex had to be removed from gimbap sass to ensure responsiveness
- Had to add transform: scale(1); to the <p> on the image to fix a weird firefox bug where the p text was disappearing when the image transitioned.  
- viewport tag added to template.js 

image

@regiskuckaertz @guardian/commercial-dev


preview/src/js/lib/template.js
```diff
@@ -8,7 +8,7 @@ let filePaths;
 let builtHTML;
 
 function buildHTML() {
-	builtHTML += '<style> body { margin: 0 } ' + filePaths.CSS.contents + '</style>' + filePaths.HTML.contents;
+	builtHTML += '<meta name="viewport" content="width=device-width, initial-scale=1">'+'<style> body { margin: 0 } ' + filePaths.CSS.contents + '</style>' + filePaths.HTML.contents;
```
Ap0c: Do you need to concat the strings here?

Calum-Campbell: No we do not. Cheers.

regiskuckaertz:
LGTM 👍  A tip for the future: it is best to favour CSS over JS, so that optional elements are always present and hidden rather than only added on an as-needed basis. For instance, instead of `showArrow` in your JS, you can just have:

```
<div class="links-with-arrows--[%linksWithArrows%]">
  <svg class="icon icon--arrow>...
</div>
```

and then in your CSS:

```
.links-with-arrows--no .icon--arrow { display: none; }
```

This would not necessarily be true in another context, but in native ads where speed is the focus we want to avoid DOM mutations.

PR #71, opened 2016-10-14
Inline LiveEvent

JonNorman:
_Finally_.

So, I use JS template strings to generate my component and thus (re)wrote `generatePicture` and `generateSources` to build the HTML from strings rather than use DOM methods.

Turns out that I could have just used the latter methods and called `outerHtml` on the object returned by `createPicture`.

Not sure which one is clearer - I'll let the PR authoriser decide! :)


JonNorman:
I haven't yet added the click-reporting to this but will do it on https://github.com/guardian/commercial-templates/pull/70 once this one is merged to master.

JonNorman:
[Inline: Live Event](https://trello.com/c/uRleu5cE/38-inline-live-event)

src/inline-liveevents/index.html
```diff
@@ -0,0 +1,15 @@
+<aside class="adverts adverts--manual adverts--legacy-inline adverts--tone-memberships advert--membership" data-link-name="merchandising | event | [%OmnitureId%]">
```
regiskuckaertz: Can you please rename `OmnitureId` `TrackingId` instead?

src/inline-liveevents/index.js
```diff
@@ -0,0 +1,49 @@
...
+.then(response => response.json())
+.then(createAdvert)
+.then(html => Promise.all([getWebfonts(), write(() => container.innerHTML = html)]))
+.then(resizeIframeHeight)
+
+function createAdvert(event) {
+
+  let lastFew = event.ratioTicketsLeft <= 0.1 ? '<div class="advert__meta advert__meta--scarcity">Last few tickets remaining</div>' : '';
+
+  return `<a class="blink advert advert--inline advert--brand" href="%%CLICK_URL_ESC%%${event.eventUrl}" data-link-name="merchandising-liveEvents-superHigh-${event.eventId}-${event.name}">
```
regiskuckaertz: I'd remove all the stuff before the event Id, it does not make sense anymore.

src/inline-liveevents/index.js
```diff
@@ -0,0 +1,49 @@
...
+        <span class="advert__more button button--small button--primary">
+            Book now ${arrowRightSvg}
+        </span>
+    </a>`;
+}
+
+function displayPriceRange(tickets) {
+
+  let prices = tickets.map(ticket => ticket.price)
+  let [low, high] = [Math.min.apply(null, prices), Math.max.apply(null, prices)]
```
regiskuckaertz: With ES6 syntax: `let [low, high] = [Math.min(...prices), Math.max(...prices)];`

src/inline-liveevents/index.js
JonNorman: Gladly, but do they not use this for tracking? I thought there were a bunch of old-looking `data-link` attributes that have stuck around because merchandise/duffet use them for reporting...?

PR #75, opened 2016-10-17
Apply fixes to Fabric Native

regiskuckaertz:
Does ~~three~~ four things:

- uses the fixed/parallax background messages for smooth scrolling effects
- uses `onViewport` to get the correct dimensions of the parent frame, rather than a media query
- adds support for animation on the second layer
- sends the maximum height for the background image in the parallax mode


Calum-Campbell:
👍 LGTM

regiskuckaertz:
[Fabric native: animated second layer](https://trello.com/c/H7RQnYDf/87-fabric-native-animated-second-layer)

PR #80, opened 2016-10-19
Corrects the case of EventID DFP variable to match that in DFP.

JonNorman:
Does what it says on the tin and HOPEFULLY SETS OFF A TEAM CITY BUILD?!?!?!!!!!1111####


regiskuckaertz:
👍

PR #81, opened 2016-10-20
Various bugs fixes and IE11 compat testing

regiskuckaertz:
A round of fixes across all templates to make them work and display correctly, especially in IE11.


regiskuckaertz:
[Native templates: round up of everything and fix bugs](https://trello.com/c/tF7vY9g9/103-native-templates-round-up-of-everything-and-fix-bugs)

regiskuckaertz:
I added an optional attribute for those rare cases where we want to set the height manually (i.e. Fabric expandables). There will be another round of tidying up the code, I'd like the writing style to be consistent and functions to be properly documented so that those things don't happen again.

PR #85, opened 2016-10-31
Add optional label to all Fabric creatives

regiskuckaertz:
As he said. The creatives now have a new attribute, `ShowLabel` that is either `yep` or `nope`. In the latter case, the label must be hidden.

picture 1

picture 2


regiskuckaertz:
Closing this PR 😞  The ad label cannot be part of the "creative space", especially if it has a fixed/parallax background:

picture 1

PR #86, opened 2016-11-01
Add optional "Advertisement" label to Fabrics

regiskuckaertz:
Yeah!

picture 1
picture 2

### DIFF
--- a/src/_shared/scss/_core.scss
+++ b/src/_shared/scss/_core.scss
@@ -127,6 +127,25 @@ $mq-breakpoints: (
     display: block;
 }
 
+.creative__label {
+    background-color: #f6f6f6;
+    color: #6e6e6e;
+    font: normal .75rem/1.25 $f-sans-serif-text;
+    height: 1.25em;
+}
+
+.creative__label--nope {
+    display: none;
+}
+
+.creative__label-text {
+    padding: 0 $gs-gutter / 2;
+
+    @include mq(tablet) {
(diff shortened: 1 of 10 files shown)


src/_shared/scss/_core.scss
```diff
@@ -127,6 +127,25 @@ $mq-breakpoints: (
     display: block;
 }
 
+.creative__label {
+    background-color: #f6f6f6;
```
JonNorman: Can you use `$neutral-8` instead?

src/fabric-video/index.scss
```diff
@@ -2,6 +2,7 @@
 
 .creative--fabric-video {
     margin: auto;
+    max-width: 1920px;
```
JonNorman: Would we benefit from storing this in a variable somewhere, e.g. `ultimate-super-wide-width-width`? :)

src/_shared/scss/_core.scss
regiskuckaertz: Yes, good call!

src/fabric-video/index.scss
regiskuckaertz: Yeah I was actually surprised this limit only applied to Fabric Video. I might use it elsewhere, so will `var` the hell out of it 😆

regiskuckaertz:
[Fabrics: add option ad label](https://trello.com/c/R8CFCcnx/120-fabrics-add-option-ad-label)

PR #92, opened 2016-11-11
Making place for AMP creatives

regiskuckaertz:
So, AMP is very specific in that:

- we can't use JavaScript: this severely limits the scope of each template to be essentially empty HTML fragments with DFP placeholders
- we can't use `style` attributes: that means moving all the style-related variables into the CSS
- AMP is for mobile only so all the desktop stuff can go away
- Finally, the `!important` keyword is forbidden in AMP... I don't know why

As a result I had to change the build process, which is the point of this PR. It also adds a Fabric for AMP for testing purposes.

(As a corollary, a much better build process will be written at some point in the future. It is good enough now. For example, we could use UNCSS to remove all the unnecessary styles for a specific template.)

### DIFF
--- a/lib/build.js
+++ b/lib/build.js
@@ -32,9 +32,19 @@ function build(options) {
             }
             prefix += `${dir}/`;
         });
+
+        let mode = dir.substr(dir.lastIndexOf('/') + 1);
+        let opts = Object.assign({ mode }, options);
+
         return [
-            buildCss(fs, `${src}/${dir}/index.scss`, `${dst}/${dir}/index.css`),
-            spawn(function*() { return buildHtml(fs, `${src}/${dir}/index.html`, yield buildJs(`${src}/${dir}/index.js`, minify), `${dst}/${dir}/index.html`, options) })
+            buildCss(fs, `${src}/${dir}/index.scss`, `${dst}/${dir}/index.css`, opts),
+            spawn(function*() { return buildHtml(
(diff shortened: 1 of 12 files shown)


regiskuckaertz:
[AMP native template: Fabric](https://trello.com/c/1nLrk5cZ/147-amp-native-template-fabric)

lib/build.js
```diff
@@ -32,9 +32,19 @@ function build(options) {
...
+        let opts = Object.assign({ mode }, options);
+
         return [
-            buildCss(fs, `${src}/${dir}/index.scss`, `${dst}/${dir}/index.css`),
-            spawn(function*() { return buildHtml(fs, `${src}/${dir}/index.html`, yield buildJs(`${src}/${dir}/index.js`, minify), `${dst}/${dir}/index.html`, options) })
+            buildCss(fs, `${src}/${dir}/index.scss`, `${dst}/${dir}/index.css`, opts),
+            spawn(function*() { return buildHtml(
+                fs,
+                `${src}/${dir}/index.html`,
+                yield buildJs(`${src}/${dir}/index.js`, opts),
```
janua: :open_mouth: :clap:

janua:
LGTM :+1:

lib/build.js
regiskuckaertz: hehe B-)

PR #93, opened 2016-11-12
Glabs Traffic Driver for apps and AMP

regiskuckaertz:
Several things going on in this PR:

- removed the `href` attribute on the `BASE` element. Since every link needs to start with the clickMacro, it did not make a lot of sense
- designed the app-specific template to match the card visual style

picture 4

- created an amp-specific template similar to the app one

picture 3

The AMP one is a POC to show that it is fairly easy to migrate an existing app template. It misses a few key features though:

- the ad roundel (because AMP lacks a popup widget and disallows the `details` element)
- the server-side fetch (because AMP disallows JavaScript)


JonNorman:
Would it be worth adding in some automatic amp validation for these templates like what we have in `frontend`?

regiskuckaertz:
Yeah I'd be happy if there was some form of validation in the future. Grab this if you want 😃  Meanwhile, DFP performs the validation so we can live with this at the beginning.

src/glabs-native-traffic-driver/amp/index.scss
```diff
@@ -72,8 +72,9 @@ $brand-colour: #69d1ca;
 
 .creative__brand-label {
     background: $brand-colour;
-    color: #767676;
+    color: #333;
```
JonNorman: Boring comment, I know, but worth sticking this in a variable since it is used elsewhere and "magic numbers" etc...?

PR #94, opened 2016-11-15
Remove max-width constraint on app traffic driver

regiskuckaertz:
Small PR, I'll give myself a +1


regiskuckaertz:
👍  Sir, your mastery of all things visual is a marvel of craftsmanship.

PR #99, opened 2016-11-19
QA fixes across all merchandising

regiskuckaertz:
Fixes all the issues observed by @ScottPainterGNM during his testing


src/_shared/scss/_adverts-jobs.scss
```diff
@@ -1,4 +1,10 @@
 .adverts--jobs {
+    @include mq($until: tablet) {
+        .adverts__row > :nth-child(2) ~ * {
```
regiskuckaertz: This is to match the same selector in `adverts.scss`, I didn't want to overthink it 😅

src/_shared/scss/_adverts-soulmates.scss
```diff
@@ -188,7 +188,11 @@
 .advert--soulmate {
     &:not(.advert--manual) {
         .advert__image-container {
-            height: 7 * $gs-baseline;
+            height: 10 * $gs-baseline;
```
JonNorman: Again, for my own edification: should this not be `max-height`?

src/_shared/scss/_adverts-jobs.scss
```diff
@@ -49,6 +55,10 @@
             background: darken($news-main-1, 5%);
         }
     }
+
+    @include mq($until: tablet) {
```
regiskuckaertz: Yeah you're probably right. The reason I did it that way is that it generates less CSS.

src/_shared/scss/_adverts-soulmates.scss
regiskuckaertz: I set the height so that usernames will align horizontally no matter what, knowing that most (if not all) profile pictures are taller than 120px. Then the image will fill its container but not more.

regiskuckaertz:
[QA: the mother of all fixes](https://trello.com/c/4xJuauHI/183-qa-the-mother-of-all-fixes)

PR #102, opened 2016-11-21
Fix misspelled class name

regiskuckaertz:
The manual inline component did not abide _by the law_ that the class names in `adverts` components must be pluralised..


regiskuckaertz:
👍  merging because I can.

PR #104, opened 2016-11-23
Fix the height of the books component and make them centred

JonNorman:
### Background
The cards within the books native component currently have no maximum height, which can lead to inconsistent looking book covers.

### Changes
This PR provides a maximum height to the cards and centres them within the image containers.

### Screenshots
**Non prominent (old)**
image

**Non prominent (updated)**
image

**Prominent (old)**
image

**Prominent (updated)**
image

### Request for feedback
@guardian/commercial-dev

### DIFF
--- a/src/_shared/scss/_adverts-books.scss
+++ b/src/_shared/scss/_adverts-books.scss
@@ -73,6 +73,44 @@
         }
     }
 
+    &.advert--prominent-true > .advert__image-container {
+        flex-basis: calc(66.67% - #{$gs-gutter / 2});
+    }
+
+    &.has-no-flex {
+        .advert--prominent-true > .advert__image-container {
+            width:  calc(66.67% - #{$gs-gutter / 2});
+        }
+
+        .advert--prominent-true > .advert__text {
+            width:  calc(33.33% - #{$gs-gutter / 2});
+        }
+    }
+
+    &.advert--prominent-false {
+      .advert__image-container {
+        align-self: center;
+        width: auto;
+      }
+      .advert__image {
+        width: auto;
+        max-height: 150px;
+      }
+    }
+
+    .adverts__row--prominent-true &.advert--prominent-false {
+      .advert__image {
+        max-height: 300px;
+      }
+    }
+
+    &.advert--prominent-true {
+      & .advert__image {
+        max-height: 450px;
+        width: auto;
+      }
+    }
+
     &.adverts--legacy-inline {
         @include mq(mobileLandscape) {
             width: gs-span(2);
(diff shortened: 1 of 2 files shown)


JonNorman:
[Fix image sizing problems in books native component](https://trello.com/c/JMnrsoC8/195-fix-image-sizing-problems-in-books-native-component)

src/_shared/scss/_adverts-books.scss
```diff
@@ -73,6 +73,44 @@
...
+      }
+      .advert__image {
+        width: auto;
+        max-height: 150px;
+      }
+    }
+
+    .adverts__row--prominent-true &.advert--prominent-false {
+      .advert__image {
+        max-height: 300px;
```
rich-nguyen: no width auto for this one?

rich-nguyen:
👍

src/_shared/scss/_adverts-books.scss
```diff
@@ -73,6 +73,44 @@
         }
     }
 
+    &.advert--prominent-true > .advert__image-container {
+        flex-basis: calc(66.67% - #{$gs-gutter / 2});
+    }
+
+    &.has-no-flex {
```
regiskuckaertz: Small mistake here. the `has-no-flex` class will be added to the `html` element so you must do `.has-no-flex &` instead. Think of the `&` as a placeholder that will unfold (through bubble up) as the compounded current selector, so:

```
.selector1 {
  /* here & means .selector1 */
  .selector2 {
    /* here & means .selector1 .selector2 */
    .has-no-flex & {
      /* the selector above will thus end up as .has-no-flex .selector1 .selector2 */
    }

    &.has-no-flex {
      /* the selector above will end up as .selector1 .selector2.has-no-flex */
    }
}
```

JonNorman: Thanks for the knowledge! As it turns out, this actually reasserts the same flex-basis that is on the flex container already so I can just remove it. It must have been left over from when I was playing around with the spacing.

regiskuckaertz: Yeah sorry, I really need to prepare that CSS workshop 😩 

Another piece of knowledge: `has-no-flex` is a **feature detection class**.  When the page load, a small script tests whether the browser supports feature X and if not, adds a class `has-no-X` on the `html` element. It means the browser does not support the `flex-*` rules and with it we can provide a workaround (here, setting a `width`).

JonNorman:
@regiskuckaertz happy for this to be merged?

regiskuckaertz:
Yip yip!

PR #105, opened 2016-11-24
Use HTTPS for client-facing links and use the bookshop link on books, not Holidays

JonNorman:
Nothing else to say here, move along.


JonNorman:
Okay - I've updated the blended, books and jobs components in DFP.

JonNorman:
[Fix native books component.](https://trello.com/c/Z4CISTMu/202-fix-native-books-component)

PR #106, opened 2016-12-02
CAPI multiple: Update mobile breakpoint design to fit the spec

regiskuckaertz:
On mobile, we should only show one article with a pic, followed by two without beneath it.

localhost-7000- iphone 6 plus 1
localhost-7000- iphone 6 plus

@ScottPainterGNM

### DIFF
--- a/src/_shared/js/capi-multiple.js
+++ b/src/_shared/js/capi-multiple.js
@@ -100,10 +100,7 @@ function buildCard (cardInfo, cardNum, isPaid) {
 
     imgContainer.insertAdjacentHTML('afterbegin', image);
 
-    // Only first two cards show on mobile portrait.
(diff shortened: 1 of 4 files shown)


ScottPainterGNM:
On apps they have gone up to 4 total as an option to replicate the desktop amount, would we be able to do the same if 4 are available?

regiskuckaertz:
Yep sure!

ScottPainterGNM:
👍  once logo is changed to top right

regiskuckaertz:
1/2 💥 

localhost-7000- iphone 6

regiskuckaertz:
localhost-7000- iphone 6 1

ScottPainterGNM:
That's the one 😻

PR #112, opened 2016-12-15
Open links in current window in merchandising components

regiskuckaertz:
As a general rule, merchandising components open links in the current window.


JonNorman:
[Native ads: open links in current window (only merch components)](https://trello.com/c/os7q0PUv/256-native-ads-open-links-in-current-window-only-merch-components)

PR #113, opened 2016-12-17
Fabric Video fix on IE

regiskuckaertz:
Check support for `object-fit` and provide alternate layout if not.


regiskuckaertz:
[Native Fabric Video: IE fixes](https://trello.com/c/wNN4TpfA/258-native-fabric-video-ie-fixes)

PR #116, opened 2017-01-09
Hosted Native Traffic Driver

lps88:
- Fix for Hosted AMP native traffic driver - background should be 5% of brand colour not yellow
- New template for web version - so we can retire the old one in `frontend`
- New js module to calculate the contrast colour from the brand colour - reduces duplication

image


lps88:
To do: automate the text colour based on brand colour contrast. We already have this mechanism in the multi-container, when it's merged in I'll pull it out into a separate module.

See: https://github.com/guardian/commercial-templates/pull/110/files#diff-82193a971c2b8792868881f4f3f68c65

src/hosted-native-traffic-driver/web/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+const OVERRIDES = {
+    headline: '[%ArticleHeadline%]',
+    imageUrl: '[%ArticleImage%]',
+    imageAlt: '[%ArticleImageAlternateText%]'
+};
+
+// Loads the article data from CAPI in JSON format.
+function retrieveCapiData() {
+    if('[%ArticleShortURL%]'.length === 0) return Promise.resolve();
```
regiskuckaertz: The precompiler (`babel`) will remove that line, because it is able to infer it will always be false. I suggest you bind the DFP var to a local:

```
let articleUrl = '[%ArticleShortURL%]';
if (articleUrl.length === 0) ...
```

src/hosted-native-traffic-driver/web/index.js
```diff
@@ -0,0 +1,60 @@
...
+
+// Uses cAPI data to build the ad content.
+function buildFromCapi ({ articleHeadline, articleUrl, articleImage }) {
+    Array.from(document.getElementsByClassName('creative__ctu')).forEach(ctu => ctu.href = articleUrl);
+
+    let title = document.getElementById('Title');
+
+    let imageContainer = document.getElementById('ImageContainer');
+    let image = generatePicture({
+        url: articleImage.backupSrc,
```
regiskuckaertz: Shouldn't you specify `OVERRIDES.imageUrl` here too?

src/hosted-native-traffic-driver/web/index.js
```diff
@@ -0,0 +1,60 @@
...
+        classes: ['creative__image'],
+        sources: !OVERRIDES.imageUrl && articleImage.sources,
+        alt: OVERRIDES.imageAlt
```
regiskuckaertz: Better ask Rob if this DFP variable is useful. He asked me to drop all the `alt` vars from three native templates.

src/_shared/js/hosted-colours.js
```diff
@@ -0,0 +1,24 @@
+import { write } from './dom.js';
+
+function isDark(hex) {
+    var colour = (hex.charAt(0) == '#') ? hex.substring(1, 7) : hex;
```
regiskuckaertz: Small thing: please use `let` or `const` instead of `var`

PR #117, opened 2017-01-10
Remove protocol from image path

regiskuckaertz:
Small fix where masterclass images were not showing up because of a dangling `${location.protocol}`.


regiskuckaertz:
👍

PR #118, opened 2017-01-11
Images downsized on manual multiple

regiskuckaertz:
Some weird behaviour in the manual multiple on Safari:
courses-images

I traced it to the fact that the container does not have a fixed `height` anymore, and so setting a `max-height` on the image is irrelevant.
picture 3


regiskuckaertz:
[Manual Multiple: images downsized on Safari on first page load](https://trello.com/c/YQxmVqCm/70-manual-multiple-images-downsized-on-safari-on-first-page-load)

PR #124, opened 2017-01-18
CAPI multiple: Add an optional brand logo to each card

regiskuckaertz:
I expect AdOps to pay due diligence and not provide both a container logo and a card logo.

picture 8

### DIFF
--- a/src/_shared/js/capi-multiple.js
+++ b/src/_shared/js/capi-multiple.js
@@ -83,13 +83,13 @@ function importCard (adType) {
 }
 
 // Constructs an individual card.
-function buildCard (cardInfo, cardNum, adType, modifyCardFn) {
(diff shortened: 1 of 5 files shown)


regiskuckaertz:
ha! I changed the following:

- remove manual override of each card logo: this was not necessary
- it turns out individual logos are already returned, I am then checking if they're all the same and if not, display logos inside the cards.

All the same:
picture 9

Different:
picture 10

regiskuckaertz:
@JonNorman yes, I noticed logos were _already_ returned for each card and we were using the last card's one. But I removed the manual override, it does not seem like a valid use case... at least for now. Instead, I introduced a test that detects when all the logos are the same or not.

JonNorman:
Ah! Awesome!

PR #126, opened 2017-01-25
Fix for Soulmates js

lps88:
Pass parameters into cleanupButtons function so that the number of DFP variables can differ.

Soulmates now only has 3 cards (since https://github.com/guardian/commercial-templates/pull/125), so trying to save the new template with `[%Offer4LinkText%]` in the js didn't work.

### DIFF
--- a/src/_shared/js/creatives.js
+++ b/src/_shared/js/creatives.js
@@ -1,7 +1,6 @@
 import { write } from './dom.js';
 
-export function cleanupButtons() {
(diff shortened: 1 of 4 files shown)


regiskuckaertz:
What about the third one, `manual-multiple`? It also calls `cleanupButtons`.

lps88:
@regiskuckaertz not sure how I missed that! 😮  Thanks!

PR #127, opened 2017-01-30
Manual multiple soulmates

lps88:
Allows user to choose between showing the fourth card and the form:

image

image

### DIFF
--- a/src/_shared/js/creatives.js
+++ b/src/_shared/js/creatives.js
@@ -1,8 +1,9 @@
 import { write } from './dom.js';
 
-export function cleanupButtons(moreTexts) {
+export function cleanupButtons() {
+    const moreTexts = ['[%Offer1LinkText%]', '[%Offer2LinkText%]', '[%Offer3LinkText%]', '[%Offer4LinkText%]'];
     const mores = Array.from(document.getElementsByClassName('advert__more'))
-    .filter((more, index) => !moreTexts[index]);
+        .filter((more, index) => !moreTexts[index]);
 
     return write(() => {
         mores.forEach(more => more.parentNode.removeChild(more));
--- a/src/manual-multiple-soulmates/test.json
+++ b/src/manual-multiple-soulmates/test.json
@@ -1,5 +1,7 @@
 {
   "ViewAll": "View all",
+  "NumberofCards": "4",
+  "IsProminent": "false",
   "Offer1LinkText": "members",
   "Offer1Image": "http://static.theguardian.com/commercial/hosted/gallery-prototype/omgb2.jpg",
   "Offer1AuthorImage": "",
--- a/src/manual-multiple-soulmates/test.json
+++ b/src/manual-multiple-soulmates/test.json
@@ -1,6 +1,6 @@
 {
   "ViewAll": "View all",
-  "NumberofCards": "4",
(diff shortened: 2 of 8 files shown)


src/manual-multiple-soulmates/test.json
```diff
@@ -1,5 +1,7 @@
 {
   "ViewAll": "View all",
+  "NumberofCards": "4",
```
regiskuckaertz: Let's ask if we can actually get rid of this attribute. It doesn't seem they used it ever.

src/_shared/js/creatives.js
```diff
@@ -1,8 +1,9 @@
 import { write } from './dom.js';
 
-export function cleanupButtons(moreTexts) {
+export function cleanupButtons() {
+    const moreTexts = ['[%Offer1LinkText%]', '[%Offer2LinkText%]', '[%Offer3LinkText%]', '[%Offer4LinkText%]'];
```
regiskuckaertz: I'm not against this change, just be aware that if one day we have a creative with more or less than 4 items, you will need to revert it 😅

src/manual-multiple-soulmates/web/index.html
```diff
@@ -56,46 +56,56 @@ <h2 class="blink__anchor advert__title">[%Offer3Title%]</h2>
                     </span>
                 </div>
             </a>
+            <a class="blink advert advert--manual advert--soulmate hide-until-tablet" href="%%CLICK_URL_UNESC%%[%Offer4URL%]" data-link-name="Offer 4 | [%Offer4Title%]" target="_top">
```
regiskuckaertz: I don't understand what this one is doing here?

src/manual-multiple-soulmates/web/index.scss
```diff
@@ -2,3 +2,11 @@
 @import '_adverts';
 @import '_adverts-soulmates';
 @import '_advert';
+
+.adverts--manual.adverts--soulmates .search {
+  margin: 0 $gs-gutter/2;
+}
+
+.adverts__row--3cards-and-form > :nth-child(4) {
+  display: none;
```
regiskuckaertz: What are you trying to achieve here?

src/manual-multiple-soulmates/web/index.html
lps88: Instead of having two separate templates I've changed the `NumberOfCards` field to allow 2 cards / 3 cards / 4 cards / 3 cards with form.

src/_shared/js/creatives.js
lps88: I know, I would rather remove it if we don't need it though 🗑  :man_dancing:

src/manual-multiple-soulmates/web/index.scss
lps88: When you want to show the form, hide the fourth card

src/manual-multiple-soulmates/web/index.html
regiskuckaertz: Right. Then I suppose you can remove `Offer4LinkText` in `cleanupButtons`, can't you?

regiskuckaertz:
Let's roll this out! In the screenshot, the grid is a bit off, just make sure it is not but otherwise LGTM 👍

PR #128, opened 2017-02-02
Update capi templates…

lps88:
- fix logo URLs in a different place since api update
- fix logic for individual logos on cards (as implemented for Paid)
- add kickers to the Paid and Supported templates (as implemented for Hosted)
- refactoring to bring the three templates in line

image
image


src/_shared/js/capi-multiple.js
```diff
@@ -10,9 +10,9 @@ const ENDPOINT = 'https://api.nextgen.guardianapps.co.uk/commercial/api/capi-mul
 
 const OVERRIDES = {
     urls: ['[%Article1URL%]', '[%Article2URL%]', '[%Article3URL%]', '[%Article4URL%]'],
+    kickers: ['[%Article1Kicker%]', '[%Article2Kicker%]', '[%Article3Kicker%]', '[%Article4Kicker%]'],
```
regiskuckaertz: These vars don't exist in CAPI multiple paid for/supported, didn't you inject them using an additional param somewhere?

src/_shared/js/capi-multiple.js
```diff
@@ -137,29 +134,32 @@ function editionLink (host, edition, adType) {
 }
 
 // Uses cAPI data to build the ad content.
-function buildFromCapi (host, cardsInfo, adType, modifyCardFn) {
+function buildFromCapi (host, cardsInfo, adType, generateLogo) {
```
regiskuckaertz: It should probably be more generic than that. Say you get a function that receives the card node and the current index as an argument:
```
function buildFromCapi (host, cardsInfo, adType, customise)`
function customise(card, index)
```

src/_shared/js/capi-multiple.js
```diff
@@ -137,29 +134,32 @@ function editionLink (host, edition, adType) {
...
+        cardList.appendChild(buildCard(info, idx, adType, cardsInfo, generateLogo));
     });
 
     return write(() => {
-        // Takes branding from last possible card, in case earlier ones overriden.
-        if( cardsInfo.isSingle ) addBranding(cardsInfo.articles.slice(-1)[0]);
+        // Takes branding from last possible card, in case earlier ones overridden.
+        var brandingCard = cardsInfo.articles.slice(-1)[0];
+        if( cardsInfo.isSingle ) {
+            addBranding(brandingCard, generateLogo);
```
regiskuckaertz: this would be more problematic. perhaps this `customise` function should be called with the whole fragment instead...

src/_shared/js/capi-multiple.js
lps88: Rob would like to add them so they're all consistent

regiskuckaertz:
Let's get this merged. I'll clean it up.

PR #129, opened 2017-02-13
Add inline book component

regiskuckaertz:
picture 4

Still need to shrink the width, but this is in frontend territory.


regiskuckaertz:
Well, the ad needs to tell gu.com that it's an inline book, so I added an optional parameter to `getIframeId` to do so. If there param is there, a `type` message will be sent to the parent frame.

regiskuckaertz:
[Native inline book](https://trello.com/c/CVE8jaJ3/150-native-inline-book)

PR #131, opened 2017-03-01
Improve the layout of paid containers

katebee:
- Add missing dev dependency (nodemon) so that `npm run preview` doesn't throw an error
- Better positioning of the paidFor label at all breakpoints

:grin: Making our natives match frontend: https://github.com/guardian/frontend/pull/16062
__________
###### BEFORE:
picture 571

###### AFTER:
picture 568

____________

###### BEFORE:
picture 572

###### AFTER:
picture 569

____________

###### BEFORE:
picture 570

###### AFTER:

picture 626

### DIFF
--- a/src/capi-multiple-paidfor/web/index.scss
+++ b/src/capi-multiple-paidfor/web/index.scss
@@ -10,6 +10,43 @@
     color: $neutral-1;
 }
 
+.advert__title {
+  @include mq($until: tablet) {
+    width: 70%;
+    min-height: $gs-baseline * 5;
+  }
+}
+
+.badge {
+  @include mq($until: tablet) {
+    text-align: center;
+    margin: 0;
+    position: absolute;
+    bottom: 0;
+    right: 0;
+  }
+}
+
+.badge__label {
+  display: inline;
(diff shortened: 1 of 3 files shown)


src/capi-multiple-paidfor/web/index.scss
```diff
@@ -10,6 +10,40 @@
...
+.advert__title {
+  @include mq($until: tablet) {
+    width: 70%;
+    min-height: $gs-baseline * 5;
+  }
+}
+
+.badge {
+  display: inline-block;
+  @include mq($until: tablet) {
```
regiskuckaertz: hmmm ... what if the width of the badge goes beyond the 30% left? I would've preferred it if you used a flex container instead, so that space math is built in.

katebee: How wide should we allow the badges to become? If we use `max-width` on the badge, and a `margin-right` on the advert-title, we can achieve the desired effect

flex is headache inducing; the image needs to be 100% width, and the badge logo is injected into the DOM by JS... 🤔

PR #135, opened 2017-03-14
Improve the layout of paid containers (with Flexbox)

katebee:
Alternative version of https://github.com/guardian/commercial-templates/pull/131. 

This one relies on flexbox support to make the title accommodate the badge. As a bonus, this version of the title is greedy and will fill the space if the badge is gone.

- Add missing dev dependency (nodemon) so that `npm run preview` doesn't throw an error
- Better positioning of the paidFor label at all breakpoints

:grin: Making our natives match frontend: https://github.com/guardian/frontend/pull/16062
__________
###### BEFORE:
picture 571

###### AFTER:
picture 568

____________

###### BEFORE:
picture 572

###### AFTER:
picture 569

____________

###### BEFORE:
picture 570

###### AFTER:

picture 626

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,6 +18,7 @@
     "express": "^4.13.3",
     "hogan.js": "^3.0.2",
     "node-sass": "^3.8.0",
+    "nodemon": "^1.11.0",
     "postcss": "^5.2.5",
     "react": "^15.3.2",
     "react-dom": "^15.3.2",
--- a/src/_shared/scss/_adverts-capi.scss
+++ b/src/_shared/scss/_adverts-capi.scss
@@ -214,6 +214,10 @@
         background: darken($paid-article-subheader, 7%);
     }
 
+    @include mq($until: tablet) {
+      flex-flow: row wrap;
+    }
+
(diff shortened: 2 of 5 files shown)


katebee:
Things I am unsure/concerned about:

What are the pitfalls of using flexbox; will it behave in all browsers we intend to support?
http://caniuse.com/#feat=flexbox

regiskuckaertz:
Yep, on a best efforts basis, we do provide alternative styles using `display: table` for instance. You can see it in the CSS surrounded by `.has-no-flex` rules.

katebee:
So this looks & behaves as expected in recent versions of Firefox, Chrome, Opera and Safari. However, I haven't been able to check IE or Edge. 😞

lps88:
I've just been trying to test this on IE on a VM, but the preview tool itself is struggling in that environment!! >.< Anyway, it looks good from here 👍

katebee:
Yes! I couldn't get the preview dropdown to work in Browserstack either... Hard to check this in IE 🤔

PR #139, opened 2017-03-23
Trackers ad nauseum

regiskuckaertz:
Adds viewability, research and impression trackers to all four Fabric styles.

todo

- [ ] test

### DIFF
--- a/src/_shared/js/ads.js
+++ b/src/_shared/js/ads.js
@@ -11,8 +11,10 @@ const GLABS_EDITION = {
     US: 'guardian-labs-us'
 };
 
-function addTrackingPixel(rootNode) {
+function addTrackingPixel() {
     if( !trackingPixel ) return;
+    loadImage(trackingPixel);
+}
 
 function addPixel(url) {
     loadImage(url);
--- a/src/_shared/js/ads.js
+++ b/src/_shared/js/ads.js
@@ -20,6 +20,14 @@ function addPixel(url) {
     loadImage(url);
 }
 
+function addViewabilityTracker(creativeId) {
+    var tracker = document.getElementById('viewabilityTracker');
+    if( !tracker ) return;
+    tracker = tracker.firstChild.nodeValue.trim();
+    if( !tracker ) return;
+    write(() => document.body.insertAdjacentHTML('beforeend', tracker.replace('INSERT_UNIQUE_ID', creativeId)));
+}
+
 function loadImage(url) {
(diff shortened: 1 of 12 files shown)


src/_shared/js/ads.js
```diff
@@ -11,17 +11,25 @@ const GLABS_EDITION = {
...
+}
 
-    if( !pixelTemplate ) {
-        pixelTemplate = document.createElement('img');
-        pixelTemplate.className = 'creative__pixel';
-    }
+function addViewabilityTracker(creativeId) {
+    var tracker = document.getElementById('viewabilityTracker');
+    if( !tracker ) return;
+    tracker = tracker.firstChild.nodeValue.trim();
```
katebee: Will `tracker.firstChild` always exist? ...and will this throw an error if `firstChild` doesn't exist?

regiskuckaertz: Very good question. The only idea I had to get this working was to use a construct such as this: `<div><!-- [%Viewabilitytracker%] --></div>`, i.e. a mini DOM tree with only one comment node containing the content of the `Viewability tracker` DFP variable. Then I can reify the content of the comment, using the method above (yes, `firstChild` will always exist, but it may be empty).

Come to think of it, is this extra step really necessary? We could instead do this in the DFP tempate:

`[%Viewabilitytracker%]`

... the only problem is the `UNIQUE_ID`. I'm not sure it is even useful. What do you think?

katebee: Hmm.... 🤔 Cool, if we are in 'control' of the DOM tree then no worries about missing children 😅 👍 

At the moment I don't really know the purpose of UNIQUE_ID. What do we use the values for later? 

IIUC: `fabric-video-expandable-` will end up with a random number appended... What then reads these? Do they end up in Ophan / GA?

JonNorman:
Shall we, er, merge this?

regiskuckaertz:
I had completely forgotten about that :rofl: are you happy with the changes? I haven't tested but it is a replica of what I did on frontend 😊

JonNorman:
Haha yeah, I'll see them out!

JonNorman:
Ignore my previous comment! As detailed in #167 , I think we can remove all pixel variables from the formats here. The tracker variables need to be retained as they are not just basic pixels.

rich-nguyen:
Closing for now, we may need this in future if a client's viewability tracker doesn't work from within an iframe.

PR #142, opened 2017-06-30
Make app-specific capi-multiple-paid-for template

kelvin-chappell:
This is very similar to the web capi-multiple-paid-for template, but without waiting to get the iframe ID.  I've tested it on iOS.


kelvin-chappell:
[GLabs merchandising (iOS/android)](https://trello.com/c/xFrqZ9PA/225-glabs-merchandising-ios-android)

kelvin-chappell:
@JonNorman Already working on the trial network.  I just copied in the html and css from the build directory.

PR #145, opened 2017-07-24
A PoC for driving traffic to GLabs content.

JonNorman:
This is a rough PoC for creating a small traffic driver ad slot.

To do:
- [x] Make the styling responsive - currently the design only services desktop in terms of width and font-size.
- [ ] Add screenshots to this PR.

### Screenshots

**Mobile**:
image

**Tablet**:
image

**Desktop**:
image

### DIFF
--- a/src/carrot/test.json
+++ b/src/carrot/test.json
@@ -1,0 +1,7 @@
+{
+  "Title": "Airbnb families",
+  "ThumbnailUrl": "https://a0.muscache.com/airbnb/static/logos/belo-200x200-4d851c5b28f61931bf1df28dd15e60ef.png",
+  "ClickthroughUrl": "https://airbnb.co.uk",
+  "Standfirst": "From Cornwall to Tuscany, plan your next adventure with insider tips and unique Airbnb homes",
+  "Tagline": "Paid for by Airbnb"
+}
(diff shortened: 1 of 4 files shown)


JonNorman:
[Get design sign-off on the Carrot.](https://trello.com/c/40181bm2/153-get-design-sign-off-on-the-carrot)

rich-nguyen:
@JonNorman I wonder if they will mention the "paid for by" label being separate from the logo. Also, does the badge style look like the whole parent article is badged?

ScottPainterGNM:
Good points by Rich, I wonder whether the more likely use case is for a link to a specific piece of content instead of the overall campaign? In which case maybe the design would be more clear?

JonNorman:
@rich-nguyen: on the logo, it wouldn't necessarily need to be a logo, I just chose that. I largely based the design off what the telegraph do (http://www.telegraph.co.uk/news/2017/07/26/princess-dianas-brother-earl-spencer-lied-princes-walking-behind/) at the top of their articles.

@ScottPainterGNM yep that is the intended use case, I think I just forgot about that along the way, I'll update the tests to point to actual content and will update this PR accordingly.

PR #149, opened 2017-08-08
Rename and redesign the Carrot.

JonNorman:
Given the images used may not be of perfect ratio, I've opted to truncate the image at smaller breakpoints so that it can still be full-height. The mockup doesn't opine either way, but some feedback would be great, or we can just give it a whirl and see what people think!

Some feedback from Hannah was that the titles/standfirsts need to accommodate longer phrases than was originally designed, hence the mobile implementation is slightly squatter/squarer than the design.

**Mockup from design:**
*Mobile*
image
*Tablet/Desktop*
image

**Preview of changes**
*Mobile*
image
*Tablet/Desktop*
image


JonNorman:
[Rework original "carrot" design to be the glabs-inline design.](https://trello.com/c/sabrnCWA/189-rework-original-carrot-design-to-be-the-glabs-inline-design)

src/glabs-inline/web/index.scss
```diff
@@ -0,0 +1,81 @@
...
+  @include mq(tablet) {
+    font-size: 12px;
+  }
+}
+
+.body__standfirst {
+  display: none;
+
+  font-size: 10px;
+  color: #333;
```
katebee: `color: $neutral-1;`

src/glabs-inline/web/index.scss
```diff
@@ -0,0 +1,81 @@
...
+}
+
+.body__tagline {
+  font-size: 10px;
+  color: $neutral-2;
+  margin-top: 4px;
+  margin-left: auto;
+}
+
+.logo__img {
```
katebee: Maybe add a small left padding/margin to the img to give it a bit of space from `paid for by`?

src/glabs-inline/web/index.html
```diff
@@ -0,0 +1,15 @@
+<aside class="container">
+  <div class="thumbnail">
+    <a href="%%CLICK_URL_UNESC%%%%DEST_URL%%">
+      <img class="thumbnail__img" src="[%ThumbnailUrl%]">
```
katebee: Can we add some alt text, or add `aria-hidden=true`? Just tried a screen reader and it reads out the src... 😅

src/glabs-inline/web/index.html
```diff
@@ -0,0 +1,15 @@
+<aside class="container">
```
katebee: `aria-label="inline advertisement"` if we want to be helpful

src/glabs-inline/web/index.html
JonNorman: Ah! Yes I will get my aria on (and also on the other PR I've just raised 😩 )

src/glabs-inline/web/index.html
JonNorman: Always!

src/glabs-inline/web/index.scss
```diff
@@ -0,0 +1,82 @@
...
+  flex-shrink: 1;
+  flex-basis: 0;
+
+  display: flex;
+  flex-direction: column;
+  justify-content: space-around;
+}
+
+.thumbnail {
+  max-width: 35%;
```
katebee: Might be worth switching to:

```
max-width: 184px;
width: 35%; 
```

Firefox sometimes gets unhappy if max-width is used without the width being set.

The value of 184px is taken from the typical trail image dimension if the height is restricted to 110px. This allows the height of the carrot to be the deciding factor, but will remove the extra whitespace from Firefox:

picture 909

Thoughts?

JonNorman: Amazing, thanks. I forgot to check firefox - great tip!

PR #153, opened 2017-08-09
Fix the widths of glabs-inline images; it makes everything simple.

JonNorman:
This PR does two things:

1. Fixes the width of the images at different breakpoints. Trying to make the image resize as the breakpoints change is a nightmare, especially considering differently proportioned images.

2. Adds `mq` breakpoint mappings that represent the width of the iframe at those breakpoints when in an article. When using media queries within an iframe, it is the size of the iframe that is queried, not the size of the device.

### DIFF
--- a/src/glabs-inline/web/index.scss
+++ b/src/glabs-inline/web/index.scss
@@ -52,7 +52,6 @@ h1, p {
   font-size: 10px;
   font-weight: bold;
   color: $neutral-1;
-  margin-bottom: auto;
 
   @include mq(articleTablet) {
     font-size: 12px;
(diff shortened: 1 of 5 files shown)


src/glabs-inline/web/index.scss
```diff
@@ -68,7 +73,7 @@ h1, p {
 .body__tagline {
   font-size: 10px;
   color: $neutral-2;
-  margin-top: 4px;
+  margin-top: auto;
```
rich-nguyen: what will this `auto` come out as?

JonNorman: Within a flex container, it sets as large a margin as possible, i.e. it sits flush against the bottom of the div.

PR #154, opened 2017-08-09
Revealer styling to match height of mpu

rich-nguyen:
A bit of styling to make sure the advertisement label is 24px height, and added a resize call too, so that the iframe height matches the content's rectangle dimensions.


JonNorman:
Do you still want this reviewing?

rich-nguyen:
oh yeah, this is still useful! Would you mind, @JonNorman ?

PR #155, opened 2017-08-14
GLabs Traffic Driver Style updates

JonNorman:
These are the final* changes to this before the test is run. They are:

1. Using a single anchor tag that overlays the entire container rather than linking individual component elements.
2. Add darken on hover, to match behaviour of inserts on frontend e.g. rich-links as well as cursor style.
3. The iframe for the glabs-left style will resize itself again. I this was removed whilst debugging a different styling issue.

*assuming we don't get drastic level of complaints about something or other.

### DIFF
--- a/src/_shared/scss/_core.scss
+++ b/src/_shared/scss/_core.scss
@@ -90,6 +90,25 @@ $creative-max-width: 1920px;
     }
 }
 
+a.u-faux-block-link__overlay {
+  position: absolute;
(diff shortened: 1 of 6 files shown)


JonNorman:
I'm going to merge this - these changes were made just before running the test to comply with the various design suggestions. Since the test has now been run, any further changes we want to make can be done in a single PR.

PR #157, opened 2017-09-07
Create a new template for wrapping around third-party creatives.

JonNorman:
A new Native style that wraps a third party script in order to provide the following behaviour:
- Provides the Guardian web fonts to the iframe
- Adds the Advertisement label
- Resizes the iframe once it is done.

It achieves this by either fetching and inserting a third party tag or by fetching the assets from a specified S3 bucket and adding the `index.html` fragment into the page.

@guardian/commercial-dev

### DIFF
--- a/src/fabric-custom/web/index.js
+++ b/src/fabric-custom/web/index.js
@@ -1,0 +1,58 @@
+//@flow
+import { getIframeId, resizeIframeHeight, getWebfonts, showAdvertLabel } from '../../_shared/js/messages';
+import { write } from '../../_shared/js/dom';
+
+const DapAssetsRoot = `https://s3-eu-west-1.amazonaws.com/adops-assets/dap-fabrics`;
+const DapAssetsFolder = '[%DapAssetsFolder%]';
+
+const DapAssetsPath = `${DapAssetsRoot}/${DapAssetsFolder}`;
+const ThirdPartyTag = '[%ThirdPartyTag%]';
+
+const ifThenPromise = (parameter, action) => {
+  if (parameter === 'true') {
+    return action();
+  } else {
+    return Promise.resolve();
+  }
+}
+
+const fetchTag = (tagUrl: string) => {
+  return fetch(tagUrl)
+    .then(response => response.text())
+}
+
+const replaceAssetLinks = (html: string) => {
+  const re = /url\('\.\/(.*)'\)/g;
+  return html.replace(re, `url('${DapAssetsPath}/$1')`);
+}
+
+const generateTag = () => {
+  if (DapAssetsFolder) {
+    return fetchTag(`${DapAssetsPath}/index.html`)
+      .then(replaceAssetLinks)
+  } else if (ThirdPartyTag) {
+      return fetchTag(`[%ThirdPartyTag%]`)
+  }
+};
+
+const insertTag = (tag) => {
+  const placeholder = document.getElementById('js-fabric-custom');
+  const range = document.createRange();
+  range.setStart(placeholder, 0);
+  range.setEnd(placeholder, 0);
+  placeholder.appendChild(range.createContextualFragment(tag));
+};
+
+const addAdvertLabel = () => {
+  const label = `<div class="ad-slot__label">Advertisement</div>`;
+  write( () => document.body.insertAdjacentHTML('afterbegin', label));
+}
+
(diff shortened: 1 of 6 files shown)


JonNorman:
[DAP - Missing Native Templates](https://trello.com/c/4BGH8C4i/88-dap-missing-native-templates)

JonNorman:
[DAP - Missing Native Templates](https://trello.com/c/4BGH8C4i/88-dap-missing-native-templates)

JonNorman:
Here is an adtest that shows the DAP template working: https://www.theguardian.com/uk?adtest=jn-native

src/fabric-custom/web/index.js
```diff
@@ -0,0 +1,58 @@
...
+  if (DapAssetsFolder) {
+    return fetchTag(`${DapAssetsPath}/index.html`)
+      .then(replaceAssetLinks)
+  } else if (ThirdPartyTag) {
+      return fetchTag(`[%ThirdPartyTag%]`)
+  }
+};
+
+const insertTag = (tag) => {
+  const placeholder = document.getElementById('js-fabric-custom');
```
regiskuckaertz: You can just do `placeholder.insertAdjacentHTML(“afterbegin”, tag)` instead of using ranges

src/fabric-custom/web/index.js
```diff
@@ -0,0 +1,58 @@
...
+  range.setEnd(placeholder, 0);
+  placeholder.appendChild(range.createContextualFragment(tag));
+};
+
+const addAdvertLabel = () => {
+  const label = `<div class="ad-slot__label">Advertisement</div>`;
+  write( () => document.body.insertAdjacentHTML('afterbegin', label));
+}
+
+const shouldShowLabel = '[%ShowLabel%]' === 'true';
```
regiskuckaertz: Look at the compiled JS to make sure the momifier did not optimise that statement away. You usually need to assign the DFP var to a JS var first

src/fabric-custom/web/index.scss
```diff
@@ -0,0 +1,17 @@
...
+    line-height: 1.25rem;
+    position: relative;
+    height: 1.5rem;
+    background-color: #f6f6f6;
+    margin: 0 auto;
+    padding: 0 .5rem;
+    border-top: .0625rem solid #dfdfdf;
+    color: #6e6e6e;
+    text-align: left;
+    -webkit-box-sizing: border-box;
```
regiskuckaertz: You can get rid of this line and the one below

src/fabric-custom/web/index.js
JonNorman: Even if the tag may contain a `<script>` tag?

src/fabric-custom/web/index.js
JonNorman: Yeah, it did originally when I was checking it directly in the `then` cascade but it seemed not to optimise it away when I added it as this singular variable.

src/fabric-custom/web/index.js
JonNorman: Sorry - I meant to ask: would the script contents be parsed and executed by the browser if I just add it using `insertAdjacentHTML`? As I understand it, I need to use the document fragment / tag pattern to achieve this.

regiskuckaertz: once you insert HTML in the DOM tree, using whatever technique (`innerHTML`, `insertAdjacentHTML` or anything really), the browser will go through all the steps to eval the new tree. so yes, if there is a `script` element somewhere, its content will be evaluated by the js engine.

JonNorman: Okay, thanks. I'll make that change and test it. Though I'm not sure where I got my original understanding from, [this article](https://ghinda.net/article/script-tags/) represents my original understanding of how this works.

regiskuckaertz: oh bummer. You're right. I did not know that :-/

regiskuckaertz: I wonder if using `createElement('script')` and then `innerText = <your js>` would work though 🤔  it works when you use the `src` attribute to download a third party script.

JonNorman: Ah yeah that might be neater. I'll check and see whether that will work with the kind of tags we're expecting :+1:

Issue #165, opened 2017-10-16
Be consistent in how the Advertisement label is added.


JonNorman:
There are a bunch of templates that add the `Advertisement` label to the top of the creative. We should define this once and call it everywhere else.

emma-imber:
Out of date - labels were removed from templates here: https://github.com/guardian/commercial-templates/pull/394

Issue #35, opened 2016-09-26
Valid test files should be provided with each native component


JonNorman:
It would make it a lot easier to check consistency of style and impact of CSS changes if each component had a valid `test.json` file, enabling it to be correctly rendered in preview.

In some cases, adding a blank `test.json` file will suffice, but in others it will result in failed fetches and thus a failure to correctly render.

annebyrne:
👍

RobertFreeman:
@JonNorman Is this still a thing you need doing? I could do with something else to practice on while I'm getting used to git...

JonNorman:
Ah! I only just saw this! Yes absolutely!

PR #166, opened 2017-10-16
A general tidy up of miscellaneous things (no YOU create separate PRs....)

JonNorman:
The main changes in this PR are:

1. When the rendering of a component depends on the response of a `fetch` call (typically to cAPI), hide the component will request to be hidden in the event of an error. In a perfect world we would have components that upgrade themselves, or at least degrade gracefully. However, this requires dev, UX and designer time, which is in short supply at the moment. This seems to be a good compromise.
**Note**: there is [a listener](https://github.com/guardian/frontend/blob/master/static/src/javascripts/projects/commercial/modules/messenger/hide.js) on `frontend` that listens for `hide` messages and hides the iframe. This isn't currently modelled in the Preview app.

2. Styling of the Advertisement label now lives in `_core.scss`. Since a lot of the different templates use this, it makes sense to pull it out into core. Note, the way the Advertisement label is added in is not ideal, and this is currently being tracked by issue #165 .

3. All the templates now have `test.json` files, thereby closing issue #35 ! The values for these were mainly taken from each template's old-version counterpart, which are usefully linked to and described [here](https://docs.google.com/presentation/d/1yASl5DP5RIGpLTLSaDFxJmbllFE0Ow4JLJ9NtPMMyvs/edit), with the new native components contrasted and described [here](https://drive.google.com/drive/u/0/folders/0B8dJervqZREuc3F1Q1dSanJLVFE) and _some_ ad tests live on the site [here](https://www.google.com/dfp/59666047#delivery/OrderDetail/orderId=2128175473) and [here](https://www.google.com/dfp/59666047#delivery/OrderDetail/orderId=348795447).

4. Small consistency amendments with camel casing. Any format variables that I've changed here, I will update in DFP and update the styles accordingly.


shtukas:
So please ignore my above approval, because I didn't mean to approve all the commits at one :)

PR #169, opened 2017-10-19
Minor fixes (proposal)

kant:


rich-nguyen:
Hi @kant , we appreciate the contribution, however we will only accept meaningful changes to the content of our readme, changes that will really help our developers contribute to the source code of the system. Actual source code contributions are even more welcome. Adding some (unbalanced) apostrophes isn't enough for us to accept the change.

PR #170, opened 2017-10-19
New Format: MPU Parallax

JonNorman:
This is a broad copy of the mobile-revealer template with a few changes.

Given the time pressure on getting this done, I will leave it to a later date to merge the two into a single template.

### DIFF
--- a/lib/css.js
+++ b/lib/css.js
@@ -4,7 +4,7 @@ const postcss = require('postcss');
 const autoprefixer = require('autoprefixer');
 const sass = require('node-sass');
 
-function buildCss(fs, src, dst, { mode = 'web', minify = undefined} = {}) {
+function buildCss(fs, src, dst, { mode = 'web', minify} = {}) {
     return spawn(function*() {
         let css;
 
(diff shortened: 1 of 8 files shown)


lib/css.js
```diff
@@ -4,14 +4,16 @@ const postcss = require('postcss');
 const autoprefixer = require('autoprefixer');
 const sass = require('node-sass');
 
-function buildCss(fs, src, dst, { mode = 'web' } = {}) {
+function buildCss(fs, src, dst, { mode = 'web', minify = undefined} = {}) {
```
JonNorman: Let's make the `minify` argument always set and a boolean.

src/_shared/js/messages.js
```diff
@@ -3,7 +3,7 @@ import { timeout } from './impl/promises';
 
 const rootElement = document.documentElement;
 
-let parentOrigin = 'http://localhost:9000';
+let parentOrigin = 'http://localhost:3000';
```
JonNorman: Is this still needed?

PR #171, opened 2017-10-20
Add tracker and pixels to Fabric custom

rich-nguyen:
Adds pixels and a viewability tracker for our Fabric custom template.


janua:
:+1:

PR #172, opened 2017-10-23
Do not evaluate constant expressions during uglify

rich-nguyen:
According to these [options]( https://github.com/mishoo/UglifyJS2#compress-options), we can disable constant expression evaluation, things like:

```
'[%ShowLabel%]' === 'true'
```

We want uglify to ignore this, because the macro will be expanded much later, inside DFP.


LATaylor-guardian:
👍

PR #175, opened 2017-11-22
Add new template for capi-multiple supported

RobertFreeman:
- New template for testing b2b article promotion
- Multiple articles 'supported by' different clients, without multiple logos

### DIFF
--- a/src/capi-multiple-nologos/web/index.scss
+++ b/src/capi-multiple-nologos/web/index.scss
@@ -1,0 +1,19 @@
+@import '_core';
+@import '_adverts';
+@import '_adverts-capi';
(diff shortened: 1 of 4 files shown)


src/capi-multiple-nologos/web/index.scss
```diff
@@ -0,0 +1,19 @@
...
+@import '_adverts-capi';
+@import '_advert';
+@import '_brandbadge';
+@import '_palette';
+
+.adverts {
+    color: $neutral-1;
+}
+
+.badge {
```
RobertFreeman: Badging <div> automatically hidden here

PR #177, opened 2017-12-04
Fix resizing in mobile revealer template

JonNorman:
This PR achieves two things:

1) Fixes a bug where the returned values from `getIframeId` were being passed into `resizeIframeHeight`, resulting in weird behaviour.

2) Removes `once` from the background update, given that `frontend` now supports refreshing a background image when the viewport changes.


LATaylor-guardian:
👍

PR #183, opened 2018-01-05
Fabrics: always set the background colour, if provided.

JonNorman:
At the moment, if a background image is not provided then the function returns without setting the background colour.

If a background image is not provided, but a background colour _is_ provided, then we should honour this. This PR alters the behaviour to reflect this.


LATaylor-guardian:
👍

PR #190, opened 2018-01-16
Update style aspects of manual-multiple template

RobertFreeman:
- Change button colour in header div and padding on blurb for manual-soulmates container


src/_shared/scss/_adverts.scss
```diff
@@ -569,12 +569,16 @@
...
+        color: #000000;
     }
 
     .adverts__blurb {
         font: normal 14px/18px $f-serif-headline;
-        color: #ffffff;
+        margin-top: 20px;
+        color: #000000;
+        background: #ffffff;
+        -webkit-box-decoration-break: clone;
```
JonNorman: Just use `box-decoration-break` here; the transpiler will add all of the necessary variants when it builds the templates.

src/_shared/scss/_core.scss
```diff
@@ -112,7 +112,7 @@ a.u-faux-block-link__overlay {
 .button {
     border: 0;
     border-radius: 1000px;
-    font: bold 13px/20px $f-sans-serif-text;
+    font: bold 13px/0px $f-sans-serif-text;
```
JonNorman: Does this not make the font-size 0?

src/_shared/scss/_adverts-books.scss
```diff
@@ -1,3 +1,8 @@
+.adverts--tone-books .button--large {
```
JonNorman: This seems to be present, unchanged, in a lot of files - is it worth factoring it out in a common file - maybe `_adverts.scss`?

JonNorman: (It doesn't seem to depend on being books, jobs etc.)

src/_shared/scss/_adverts.scss
RobertFreeman: I'll change that - sorry, you told me that already...

src/_shared/scss/_core.scss
RobertFreeman: It makes the line height 0px - I did that to vertical-align the button text in the centre (the studio said the text was too high and mis-aligned with the arrow), sure there's a better way of doing that.

src/_shared/scss/_adverts-books.scss
RobertFreeman: Definitely that would be much cleaner - I'm going to try and refactor the code in the next couple of weeks, hopefully remove unnecessary files and code...

src/_shared/scss/_core.scss
JonNorman: Hmm, what is the content of the button? If there isn't any text then presumably we can just set `line-height: 0`; maybe we can look at this together and see if there is a cleaner way of achieving this (if `line-height` doesn't work)

RobertFreeman: I just checked and we can just delete the line-height bit altogether I think and that'll do the same thing. The button small div is:

image

JonNorman: Nice one, deleting code is always a bonus 😃

PR #191, opened 2018-01-17
Force styles to explicitly request/refuse an Advertisement label.

JonNorman:
Now that https://github.com/guardian/frontend/pull/18924 has been merged, we need to update the styles here to explicitly say whether they request/refuse an Advertisement label.

This will make the logic in `frontend` a lot simpler and will mean that no creatives need to include their own Advertisement label. Huzzah.

### DIFF
--- a/lib/html.js
+++ b/lib/html.js
@@ -3,7 +3,7 @@ const hogan = require('hogan.js');
 
 function buildHtml(fs, src, js, dst, { banner = '', footer = '', mode = 'web' } = {}) {
     const partials = {
-        svg: () => (text, render) => {
+        svg: () => text => {
             const svgName = text.trim();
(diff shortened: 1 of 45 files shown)


lib/html.js
```diff
@@ -17,6 +17,14 @@ function buildHtml(fs, src, js, dst, { banner = '', footer = '', mode = 'web' }
             }
         },
 
+        // this is used by frontend to determine whether an advertisement label is required
+        // the presence of this div will indicate that no such label should be rendered
+        labelRequired: () => (text, render) => {
```
rich-nguyen: what does `render` do?

JonNorman: Absolutely nothing. Removing it.

PR #195, opened 2018-02-09
Rf garnett changes round2

RobertFreeman:
This is the second round of cosmetic changes for the styling of commercial merchandising components, fixes for:

- Update GLabs single container (for Australia) and manual inline logos (for consumer revenues - GLive, GBookshop, GMasterclasses)

- Update Guardian Bookshop component with new header div (inc. blurb) and left-align book images within container

- Changed flex position and direction so on mobile, CTA buttons and logo were positioned correctly

- Changed colours and font styles for capi-supported container (to match current site w/o any blue!)


src/_shared/scss/_adverts-capi.scss
```diff
@@ -30,6 +30,9 @@
 .adverts--paidfor {
     .adverts__header {
         position: relative;
+        -webkit-flex-direction: column;
```
JonNorman: You only need to provide the `flex-direction` property here; the compilation from SCSS to vanilla CSS adds in all the engine-specific alternatives.

RobertFreeman: done!

PR #201, opened 2018-06-01
Create a Fabric template to serve third party tags

frankie297:


src/fabric-third-party/web/index.js
```diff
@@ -0,0 +1,36 @@
+//@flow
+import { getIframeId, resizeIframeHeight, getWebfonts } from '../../_shared/js/messages';
+import { write } from '../../_shared/js/dom';
+
+const ThirdPartyTag = '[%ThirdPartyTag%]';
```
katebee: We don't seem to use this const? 

Also JS convention of using lower camel case when naming would probably make this:
`const thirdPartyTag = ...`

PR #208, opened 2018-09-27
Add new styles to manual multiple template 

RobertFreeman:
This is a PR to update the manual multiple template with Guardian Weekly and Guardian Members branding

- Updated flex-direction in manual .scss to fix alignment of logo/button across tablet/mobile breakpoints
- Added 2 new .scss files (weekly + members)
- Added 2 new .svg files (weekly + members)
- Updated js function in template html to import new logos
- Added 2 new hex codes in palette.scss (weekly + members)

desktop
tablet
mobile


src/_shared/scss/_adverts-members.scss
```diff
@@ -0,0 +1,57 @@
+.adverts--memberss {
```
katebee: `memberss` ?? Why the second s? 

Follow up question - Where do these classnames come from?

RobertFreeman: Fair question - literally the class names on these are so annoying, and I don't know why the second s, this code has been through 3 different engineers in 2.5 years and it is weeeiiird. I had ambitions to make it my life mission to clean everything up and consolidate loads of the css, but sadly I wasn't allowed.

katebee: Copying over some internal discussion to help any future git history explorers 😁 

The classes are inserted into the HTML of the template here:
https://github.com/guardian/commercial-templates/blob/master/src/manual-multiple/web/index.html

> the tone ('members' 'travel' 'live' etc.) comes from a DFP value, then for some reason there is an extra 's'
> Which means the class names are memberss, travels, lives
> It's a mystery to this day

😸 😸 Okay! Cheers for the explaination 👍

RobertFreeman: Thanks Kate!

PR #209, opened 2018-10-08
Fix broken util task by updating DFP API references

katebee:
DFP is now called AdManager...

This breaks a lot of things in the API since the version on which we were relying is now sunsetted.

This change will make the StyleSync tool work once more!

Thank the fabulous @JonNorman for the fix 🙏 🙇 🎉

### DIFF
--- a/style-sync/build.sbt
+++ b/style-sync/build.sbt
@@ -5,7 +5,7 @@ scalaVersion := "2.12.3"
 val slf4jVersion = "1.7.5"
 
 libraryDependencies ++= Seq(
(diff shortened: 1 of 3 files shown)


JonNorman:
Looks good! I don't think I can approve things in this repository anymore... 😭

JonNorman:
@kelvin-chappell it was the start of an App that could automatically keep DFP and our Commercial Templates in sync. I only got as far as it reporting the differences - but it wouldn't be too much work to get it to push local changes up to DFP.

PR #211, opened 2018-10-30
Change carrot brand logo size and standfirst line height

frankie297:
Does what it says on the tin.

### DIFF
--- a/src/carrot/web/index.scss
+++ b/src/carrot/web/index.scss
@@ -46,7 +46,7 @@ h1, p {
 
 .body__title {
   font-size: 1.25rem;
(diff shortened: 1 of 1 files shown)


RobertFreeman:
Looks great to me Frankie - only thing is it's now so big the 'paid for by' text does look a bit weird being bottom-aligned, not a deal-breaker at all though.

image

frankie297:
Have also changed the font-weight of the title to "normal" from "bold". There doesn't seem to be an inbetween, it's either extremely bold or light. 

Before:
picture 146

After:
picture 147

@RobertFreeman @katebee

PR #200, opened 2018-05-15
First iteration of DFP Epic template

Mullefa:
Add the first iteration of the DFP Epic template. The corresponding Native Template in DFP is named _Epic V1_.

__Comparison between Native and DFP Epic templates:__

|             | Native                                                                                                                 | DFP                                                                                                                 |
|-------------|------------------------------------------------------------------------------------------------------------------------|---------------------------------------------------------------------------------------------------------------------|
| __Desktop__ | native_desktop | dfp_desktop |
| __Mobile__  | native_mobile  | dfp_mobile  |

__Validation that the link is working correctly:__

Screen grab of the (relevant) part of the redux store on Support Frontend after clicking the _Support The Guardian_ button on the DFP Epic.

redux_state

__Implementation note:__

The CSS was derived from grabbing the relevant snippets of `scss` from frontend. The resultant file could be simplified but unless it's actually incorrect, I'd prefer to delegate this work to a separate pull request (for expediency; this template is going to be used in a frontend test we want to start running ASAP).

__General changes__

Some tweaks to the 'global' code in the repository were required when implementing this template:

- consider `%%PATTERN:variable%%` as a placeholder
- add `URL` to the list of function polyfilled
- change `f-serif-text` to same value as defined in Guardian frontend
- change `multi-media-2` to same value as defined in Guardian frontend
- add colour `$news-garnett-highlight`

PR #212, opened 2018-10-31
Update CAPI templates after new dotcom fonts

RobertFreeman:


src/_shared/scss/_core.scss
```diff
@@ -1,6 +1,6 @@
 $f-serif-headline: 'Guardian Egyptian Web', Georgia, serif !default;
 $f-sans-serif-headline: 'Guardian Text Sans Web', 'Helvetica Neue', Helvetica, Arial, 'Lucida Grande', sans-serif !default;
-$f-serif-text: 'Guardian Text Egyptian Web', Georgia, serif !default;
+$f-serif-text: 'Guardian Egyptian Web', Georgia, serif !default;
```
katebee: Why this change, I thought the font names were the same? 🤔 

Looks like `Guardian Egyptian Web` got changed to `Guardian Text Egyptian Web` with PR #200 in relation to the Epic work... Do we need separate variables so the different templates all play nice?

PR #215, opened 2019-02-05
Add tracking pixel img to capi template HTML

RobertFreeman:
Adding an  tag with the class name creative__pixel to the bottom of the HTML in the capi-multiple template.

(also updated the test.json file with live microsite).

### DIFF
--- a/src/capi-multiple-paidfor/test.json
+++ b/src/capi-multiple-paidfor/test.json
@@ -1,10 +1,10 @@
 {
-    "ComponentTitle": "Barclays Lets Go Forward",
-    "SeriesURL": "barclays-lets-go-forward",
+    "ComponentTitle": "Sweet Veganuary",
+    "SeriesURL": "sweet-veganuary",
     "Brand": "Barclays",
     "TrackingID": "",
     "Article1Headline": "",
-    "Article1Kicker": "GDPR",
+    "Article1Kicker": "Vegans",
     "Article1URL": "",
     "Article1Image": "",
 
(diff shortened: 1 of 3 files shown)


src/capi-multiple-paidfor/test.json
```diff
@@ -1,10 +1,10 @@
 {
-    "ComponentTitle": "Barclays Lets Go Forward",
-    "SeriesURL": "barclays-lets-go-forward",
+    "ComponentTitle": "Sweet Veganuary",
+    "SeriesURL": "sweet-veganuary",
     "Brand": "Barclays",
```
katebee: Minor: this is probably irrelevant since file is for local testing of layouts

Did you want to also change this line to update the brand too?

src/capi-multiple-paidfor/test.json
```diff
@@ -1,10 +1,10 @@
...
     "TrackingID": "",
```
katebee: You might want to add a key into the test json too:

```json
  "TrackingPixel": "",
```

katebee: This is not necessary though! But it may allow you to test adding the pixel 😄

src/capi-multiple-paidfor/test.json
RobertFreeman: All done (in the other commit)!

src/capi-multiple-paidfor/test.json
RobertFreeman: This too...

PR #219, opened 2019-04-30
fix to allow moat tag to work within creative, by removing div - shor…

Onelittlewolfe:
fix to allow moat tag to work within creative, by removing div


katebee:
@Onelittlewolfe are you happy to merge this? 👍

PR #224, opened 2020-01-14
Add third party pixels to Interscroller Ad

ioanna0:
- Add third party pixels to Interscroller Ad
- Switch using GAM macro for destination URL


Onelittlewolfe:
Looks good.

PR #232, opened 2020-02-20
Fix truskins on slow networks

ioanna0:
This is a fix for truskins which went out of sync on slow networks. The truskin was not working as expected because the fabric & the truskin weren't loaded at the same time. Sending an event to the frontend from the commercial template makes sure both are rendered before we do the calculations for repositioning them

Co-authored-by: Josh Buckland <buck06191@gmail.com>

### DIFF
--- a/src/truskin-page-skin/web/index.js
+++ b/src/truskin-page-skin/web/index.js
@@ -1,15 +1,14 @@
-(function () {
-    let parentBody = window.parent.document.body;
-    parentBody.style.backgroundImage='url(%%VIEW_URL_UNESC%%[%PageSkinFileName%])';
-    parentBody.classList.add('truskin-page-skin');
-    parentBody.style.backgroundAttachment='fixed';
-    parentBody.style.backgroundPosition='50% 0';
-    parentBody.style.backgroundRepeat='no-repeat no-repeat';
-    parentBody.onclick=function(e) {
-        const target = e.target;
-        if (target.nodeName.toLowerCase() === 'body') {
-            window.open('%%CLICK_URL_UNESC%%[%ClickDestination%]');
-        }
-        return true;
-    };
-})();
+let parentBody = window.parent.document.body;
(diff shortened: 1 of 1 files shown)


src/truskin-page-skin/web/index.js
```diff
@@ -1,15 +1,14 @@
...
+parentBody.style.backgroundRepeat='no-repeat no-repeat';
+parentBody.onclick=function(e) {
+    const target = e.target;
+    if (target.nodeName.toLowerCase() === 'body') {
+        window.open('%%CLICK_URL_UNESC%%[%ClickDestination%]');
+    }
+    return true;
+};
+
+window.top.postMessage('truskinRendered', '*');
```
GHaberis: Could we be call `window.parent.postMessage` instead of `window.top.postMessage`? 

`window.top` targets the top level window when there's a nested hierarchy of window objects, whereas window.parent explicitly targets the immediate parent.  We're getting `parentBody` from `window.parent` so I assume this is setup to work with the immediate parent only. You might not get the expected behaviour if for some reason there's a hierarchy of window objects greater than 2, I don't know how likely that is though!

ioanna0: Good point. We want it to target the `top` window in cases when there are nested iframes so we have now switched on using `window.top.document.body` instead of `window.parent.document.body`

buck06191: Following this change, I think this can be merged now?

src/truskin-page-skin/web/index.js
```diff
@@ -1,15 +1,14 @@
...
-    parentBody.style.backgroundRepeat='no-repeat no-repeat';
-    parentBody.onclick=function(e) {
-        const target = e.target;
-        if (target.nodeName.toLowerCase() === 'body') {
-            window.open('%%CLICK_URL_UNESC%%[%ClickDestination%]');
-        }
-        return true;
-    };
-})();
+let parentBody = window.top.document.body;
```
buck06191: Looks good

PR #236, opened 2020-03-06
Native traffic fix for australia

buck06191:
Native traffic drivers were failing to resize in Australia. This seems to be because an explicit return statement was required as is now seen on [line 110 of `src/_shared/js/messages.js`](https://github.com/guardian/commercial-templates/compare/native-traffic-fix-for-australia?expand=1#diff-464152bc8d3f5065a5d49485bb558a3bR110).

Adding this in appears to have fixed the problem, which has been tested using browserstack against a machine in Australia as well as against UK machines.

### DIFF
--- a/src/_shared/js/messages.js
+++ b/src/_shared/js/messages.js
@@ -106,8 +106,12 @@ function getWebfonts(fontFamilies) {
 function resizeIframeHeight(height = -1) {
     return height === -1 ?
         timeout(Promise.all(areImagesLoaded().concat(isDocumentLoaded())), 3000)
-        .then(() => read(() => document.body.getBoundingClientRect().height))
+        .then(() => {
+            console.log("*** Images loaded in under 3 seconds")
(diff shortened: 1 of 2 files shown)


src/_shared/js/messages.js
```diff
@@ -106,11 +106,14 @@ function getWebfonts(fontFamilies) {
...
-        .then(() => read(() => document.body.getBoundingClientRect().height))
+        .then(() => {
+            return read(() => document.body.getBoundingClientRect().height)
+        })
         .then(function(height) {
+            console.log(`*** Sending resize height of ${height}`);
             return sendMessage('resize', { height });
         }) :
-        sendMessage('resize', { height });
+        (console.log(`*** Height !== -1. Sending resize height of ${height}`), sendMessage('resize', { height }));
```
ioanna0: let's remove the console logs for the native template we can still use the test one for debugging

PR #239, opened 2020-03-18
Remove redundancy in the code base

buck06191:
The masterclass and live events automated adverts were effectively using
the `[%Tone%]` to determine styling but had instead had this hard coded
into each advert.
This moves to using the `[%Tone%]` variable directly, reducing redundancy
in the code base and making the adverts easier to implement in Google Ad
Manager.

Also updates the `test.json` to use the correct text for the example given.

## After change: 
image
Styling remains the same


src/events-multiple/web/index.scss
```diff
@@ -1,5 +1,22 @@
 @import '_core';
 @import '_adverts';
+@import '_adverts-books';
```
ioanna0: Do we need these css's to allow different styles? I assume yes if this is no longer masterclass specific but double checking

buck06191: I was going to remove them because we're only currently supporting masterclass and live events, but I think that the impact to the advert is minimal and it should make updating in the future much easier.
Happy to remove if we decide otherwise.

PR #243, opened 2020-06-10
Remove style-sync

buck06191:
## What does this change?
`style-sync` is not in use and has a number of security vulnerabilities flagged up by Snyk. By removing it we reduce the need to maintain code that is not in use. If needed again, we can revisit and bring it back from git history.


buck06191:
https://github.com/guardian/commercial-templates/pull/243/commits/ec65ad366e5dd494c4186faffce52667d5a89e68 was pushed to trigger the checks after removing style-sync from Snyk

PR #248, opened 2020-11-24
Sw viewabiltiy tracking issue

Onelittlewolfe:
## What does this change?
For the Interscroller format the creative is injected further up the page via frontend and the iframe height was not been set, which was causing an issue with the activeView impression firing.

## How to test
Test format setup at 

https://admanager.google.com/59666047#creatives/custom_native_style/detail/styleid=400193

Adtest on production
https://www.theguardian.com/fashion/2020/nov/04/put-my-own-makeup-on-for-the-first-time?dcr=false&adtest=ArmaniEAGent2

Should be able to see the activeView event firing.

Needs to be tested on mobile devices, but I can't see any potential risks.


ioanna0 (approved):
👍

PR #254, opened 2021-04-14
Msav/svg dedup

MarSavar:
## What does this change?
Both the `guardian-generic-logo` and `guardian-subscriptions-logo` SVGs were duplicated several times in  `manual-multiple` template (the generic logo was loaded five times and the subscriptions logo twice). 
This change simplifies the code so the build size of is reduced from 87KB to 64KB (26% reduction).

More succinct code, smaller size.

## Images
No visual changes, but I'm adding screenshots for both logos to show that this change doesn't affect the rendering of the SVGs. Examples:

`subs-rebrand` tone

image

`climate` tone

image

### DIFF
--- a/src/manual-multiple/web/index.html
+++ b/src/manual-multiple/web/index.html
@@ -71,10 +71,7 @@
         </div>
     </div>
 </aside>
-<script>
-    const genericLogoTones = ['brand', 'lifestyle', 'climate', 'climate2', 'support']
-    const subscriptionsLogoTones = ['subscription', 'subs-rebrand']
-    
+<script>    
     const genericLogoString = '{{#svg}}guardian-generic-logo{{/svg}}'
     const subscriptionsLogoString = '{{#svg}}guardian-subscriptions-logo{{/svg}}'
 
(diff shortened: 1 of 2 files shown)


mxdvl (commented):
Did we run a test with sample values? It would be worth adding a screenshoot.

MarSavar:
Yes, I tested all the tones locally for both the generic logo and the subscriptions one.
I didn't add screenshots because there's no before/after change, but I agree, it's worth showing that this change doesn't break anything so I've edited my original PR comment. Thanks for the tip!

src/manual-multiple/web/index.html
```diff
@@ -72,22 +72,23 @@ <h2 class="blink__anchor advert__title">[%Offer4Title%]</h2>
...
         'weekly': '{{#svg}}guardian-weekly-logo{{/svg}}',
         'members': '{{#svg}}guardian-members-logo{{/svg}}',
         'patron': '{{#svg}}guardian-patron-logo{{/svg}}',
-        'lifestyle': '{{#svg}}guardian-generic-logo{{/svg}}',
-        'climate': '{{#svg}}guardian-generic-logo{{/svg}}',
-        'climate2': '{{#svg}}guardian-generic-logo{{/svg}}',
-        'support': '{{#svg}}guardian-generic-logo{{/svg}}',
-        'subs-rebrand': '{{#svg}}guardian-subscriptions-logo{{/svg}}',
+        ...Object.fromEntries(genericLogoTones.map(tone => [tone, genericLogoString])),
+        ...Object.fromEntries(subscriptionsLogoTones.map(tone => [tone, subscriptionsLogoString]))
```
mxdvl, line 92: Considering there’s [no support for Object.entries in IE][caniuse], we could also simply make it a clean object like so:

```suggestion
        'brand': genericLogoString,
        'lifestyle': genericLogoString,
        'climate': genericLogoString,
        'climate2': genericLogoString,
        'support': genericLogoString,
        'subscription': subscriptionsLogoString,
        'subs-rebrand': subscriptionsLogoString,
```

[caniuse]: https://caniuse.com/?search=Object.entries

MarSavar: I hadn't thought about compatibility - you're absolutely right. 
The build size is also not affected by this change, which is great, and it's also much better in terms of code readability.

ioanna0 (approved):
Much cleaner 👍

PR #256, opened 2021-04-30
node-sass is deprecated and is being replaced with dart-sass 

Onelittlewolfe:
## What does this change?
[node-sass](https://www.npmjs.com/package/node-sass) is used as a dev dependency, but is now deprecated and not supported. This request swaps that dependency to sass which now uses dart-sass, which is compatible with node-sass.

https://www.npmjs.com/package/sass

sass.renderSync appears to be fully supported
https://github.com/sass/dart-sass/blob/master/README.md#javascript-api

We use sass.types here.
https://github.com/guardian/commercial-templates/blob/main/lib/css.js#L34

I couldn't see anything explicit within dart sass docs to suggest this was or wasn't supported. Testing locally showed no errors though and still functioned ok.

## How to test
Run npm i
Run npm run preview or npm run build

Previews should render all ad previews with no change and build script should build minified builds with no errors.

### DIFF
--- a/lib/css.js
+++ b/lib/css.js
@@ -2,7 +2,7 @@ const spawn = require('./task');
 const csso = require('csso');
 const postcss = require('postcss');
 const autoprefixer = require('autoprefixer');
-const sass = require('node-sass');
+const sass = require('sass');
 
 function buildCss(fs, src, dst, { mode = 'web', minify} = {}) {
     return spawn(function*() {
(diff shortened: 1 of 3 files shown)


MarSavar:
Tested it locally, didn't spot any changes in the preview or errors when building.
On a separate note, `npm audit` reveals 3 vulnerabilities caused by `yargs-parser` and `node-fetch`, which are unrelated to `dart-sass` (I think), but I thought I'd flag it here just in case.

mxdvl (approved):
Love it, that means we can bump the node version, too!

PR #257, opened 2021-05-06
Msav/meta text colour

MarSavar:
## What does this change?
This changes the colour of meta and offer text from 
image (#767676) light grey 
to 
image (#121212) dark grey
in the single and multiple merchandising templates.

This is so the contrast check passes, as the light grey tone didn't have enough contrast against the background.

## Images

`manual-single` before

image
`manual-single` after

image

`manual-multiple` before

image
`manual-multiple` after

image

## Accessibility

- [x] [Colour contrast passed](https://www.theguardian.design/2a1e5182b/p/6691bb-accessibility/t/92b913)

### DIFF
--- a/src/_shared/scss/_advert.scss
+++ b/src/_shared/scss/_advert.scss
@@ -101,6 +101,7 @@
     .advert__standfirst {
         margin-top: $gs-baseline / 2;
         opacity: .7;
+        color: $guardian-brand-dark;
     }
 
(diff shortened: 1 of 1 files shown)


mxdvl (approved):
Great!

We're doing a check with `wcag-contrast` in @guardian/libs, so it may be worth using it here, too, if we ever refactor.

PR #267, opened 2021-08-18
Update native ad header alignment

chrislomaxjones:
## What does this change?

This PR fixes a layout issue where the header of native adverts in merchandising slots does not align with sections on the rest of the website. 

The fix is to increase the the width of the header by half a gutter width (and an additional pixel to account for the border drawn by other sections). Additionally the spacing in the advert row and body has been adjusted to account for the change in width of the header.

| Before      | After      |
|-------------|------------|
| Screenshot 2021-08-19 at 14 47 44 | Screenshot 2021-08-19 at 14 47 54 |

Note that this does not affect paid for adverts delivered through frontend.

## How to test

Example test ads can be viewed at:
- https://www.theguardian.com/uk?adtest=native-alignment-test (For manual multiple and capi-multiple-paidfor)
- https://www.theguardian.com/uk?adtest=native-alignment-test-single (For manual single)

### DIFF
--- a/src/_shared/scss/_adverts.scss
+++ b/src/_shared/scss/_adverts.scss
@@ -1,12 +1,13 @@
 @mixin separator($colour, $left, $margin-left: 0) {
     content: '';
     position: absolute;
-    left: $left;
+    left: 0;
     top: 0;
     bottom: 0;
     margin-left: $margin-left;
     width: 1px;
     background: $colour;
+    box-sizing: border-box;
 }
 
 .adverts {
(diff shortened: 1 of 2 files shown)


src/_shared/scss/_adverts.scss
```diff
@@ -66,7 +66,8 @@
         > .adverts__header {
             box-sizing: border-box;
             padding: $gs-baseline $gs-gutter / 5 * 2;
-            width: 160px;
+            // Extra 1px to line up with borders on sections above / below
+            width: 160px + $gs-gutter/2 + 1px;
```
arelra, line 70: Not a biggie but it would be nice to have consistent formatting e.g. `160px + $gs-gutter / 2 + 1px`

arelra (approved):

PR #268, opened 2021-08-19
Remove tracking id from Capi Multiple Paid For templates

chrislomaxjones:
## What does this change?
Remove the `[%TrackingId%]` from CAPI Multiple Paid For native ads.

## Why

This parameter is not provided via the native format in GAM - this causes an error when attempting to publish the template style since it expects a Tracking Id and none is provided.


ashishpuliyel (approved):
I notice this repo doesn't actually require approval (we should probably fix that), but here you go anyway.

chrislomaxjones:
> I notice this repo doesn't actually require approval (we should probably fix that)

Have just added some simple branch protection to main

PR #269, opened 2021-09-29
Fix meta text for Events Multiple templates

chrislomaxjones:
## What does this change?

Move the meta text associated with each card from a child of the title text to sibling. This has the effect of not applying header styling to the meta text.

## Why

Meta text shouldn't have the advert title styling applied. If this is desired the text can be included as part of the title from within Ad Manager.

### Before
Screenshot 2021-09-29 at 10 09 35

### After
Screenshot 2021-09-29 at 10 44 45


arelra (approved):

mxdvl:
This is definitely not Bimini Bon Boulash:

not bimini

PR #272, opened 2021-12-03
Allow 1 element in hosted multiple

mxdvl:
- allow a single-element
- bump node
- upgrade package-lock
- style one element correctly


mxdvl:
Nevermind… we need to modernise this first! #275

PR #273, opened 2021-12-07
Bump dependencies

mxdvl:


mxdvl:
This is just too hard and annoying… we should simply use another framework

PR #274, opened 2021-12-07
Update CAPI single paidFor

mxdvl:
## What does this change?

Update the `capi-single-paidfor` template on US’s request.

Compared to the previous version, the image is more prominent, and the badge link is removed.

## How to test

`npm run preview`, or via an `adTest` param and GAM.

Better engagement.

## Images

capi-single-paidfor template


src/capi-single-paidfor/web/index.js
```diff
@@ -59,25 +53,18 @@ function populateCard(host, responseJson) {
...
-      ${arrowRight}
-    </a>
-    </div>
     <div class="badge js-badge">
       Paid for by
       <a class="badge__link" href="${clickMacro}${responseJson.branding.logo.link}" data-link-name="badge">
         <img class="badge__logo" src="${getValue('[%BrandLogo%]', responseJson.branding.logo.src)}" alt="">
       </a>
+    </div>
     </div>`;
```
arelra, line 69: Minor if you want to - can we improve the HTML indentation for this block?

mxdvl: nah

mxdvl: Go on, then…

src/capi-single-paidfor/web/index.scss
```diff
@@ -1,15 +1,57 @@
...
+      flex-grow: 1;
+  }
+
+  .badge {
+    width: 200px;
+    text-align: left;
+
+    &::before {
+      display: none;
+    }
```
arelra, line 45: This seems a bit weird? Or am I missing some CSS trick?

mxdvl: You _are_ missing a trick: it's SASS.

The `&` is replaced by the parent item when converted to CSS. Here we could simplify it if we're not adding any other rule…

arelra: Yeah I understand its SASS but why the `display:none` on `::before` ?

mxdvl: It's overriding a rule that displays a line…

zekehuntergreen (approved):
👍

PR #275, opened 2021-12-14
Use SvelteKit to build templates

mxdvl:
## What does this change?

This is a complete rewrite of how we build commercial templates. We have found it harder and harder to work on these templates, as they are seldom updated and it’s hard to understand what each template looks like in practice. The ever-growing amount of CSS makes it hard to make changes confidently. By using components and making the templates be deployed on their own microsite, we can solve these two issues.

Development of the templates is also improved, as templates can be continuously rebuilt whilst working on them.

In terms of technology, this uses Svelte components for building template elements. SvelteKit is how we have a preview environment, that we can publish on github pages as a static site.

- [X] Hot reloading
- [X] Deployment to GH Pages (not automated yet)
- [X] Real Dynamic (CSR) example (`capi-single-paidfor`)
- [x] Real Static (SSR) example (outstanding: include some JS)
- [ ] Typings of iframe messenger

All the the previous files are kept in `legacy`.

## How to test

`npm run dev`

## How can we measure success?

Easier to work with, live reloading, modern tooling.

The resulting code is no larger that the existing one, and performs as fast.

## Have we considered potential risks?

There’s a lot of knowledge that went into the original templates. It’s possible that new templates initially lack some of the features. This needs to be thoroughly tested before being put to production.

## Images

image

PR #276, opened 2021-12-15
Move all the templates to `legacy`

mxdvl:
## What does this change?

Paves the way for #275. By moving all the existing templates to the `legacy` folder, we can keep both codebases around, rather than starting from a completely fresh point.

## How to test

History is kept through `blame`, as you can see here:

https://github.com/guardian/commercial-templates/blame/svelte-kit-migration/legacy/src/_shared/scss/_advert.scss

We keep the legacy code, if we need to refer to it, whilst preparing the way for the new SvelteKit codebase.

You cannot use the commit history without blame, because of git’s limitation with renames. You can still do this locally with the `--follow` flag, like so:

```sh
# You can use any file for $filename
filename=legacy/src/manual-multiple/web/index.scss &&\
git log --pretty=oneline --follow $filename
```


chrislomaxjones (approved):
Looks good!

mxdvl:
You’ll have to merge yourself… or give write access to `@guardian/guardian-developers-write`

PR #295, opened 2022-07-21
Consolidate plugins in eslint config

SiAdcock:
## What does this change?

Consolidate plugins property in `.eslintrc.cjs` – there are currently two entries for this property, which causes ESLint to fail with the message:

> invalid: 'svelte3/svelte3' was not found

For convenience, I've also made the following changes:

- added a `lint:fix` script that automatically fixes lint errors detecting to ESLint, if possible
- ignored `legacy/` code from ESlint run – mainly because it finds so many errors, and fixing them seems lower priority right now. Happy to discuss!

## How to test

Run ESLint:

```sh
$ npm run lint
```

Observe that the error message no longer appears and ESLint run completes successfully – albeit with legitimate linting errors we can fix later.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -2,8 +2,8 @@ module.exports = {
 	root: true,
 	parser: '@typescript-eslint/parser',
 	extends: ['@guardian/eslint-config-typescript'],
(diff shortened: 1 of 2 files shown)


arelra (approved):
Great to see the templates getting some TLC

PR #298, opened 2022-07-26
Upgrade sveltekit and vite

SiAdcock:
## What does this change?

- upgrades to the latest versions of `@sveltejs/kit` and `vite`. This exposes vite to the developer, as opposed to hiding it behind the svelte-kit CLI.

## How to test

The following scripts still work:

```sh
$ npm run dev
$ npm run build
$ npm run preview
```

### DIFF
--- a/svelte.config.js
+++ b/svelte.config.js
@@ -1,6 +1,5 @@
 import adapter from '@sveltejs/adapter-static';
 import preprocess from 'svelte-preprocess';
-import { defineConfig } from 'vite';
 
 const isDev = process.env.NODE_ENV === 'development';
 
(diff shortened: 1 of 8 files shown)


svelte.config.js
```diff
@@ -13,51 +12,16 @@ const config = {
...
 
 		// Github pages really likes its trailing slashes!
 		trailingSlash: isDev ? 'never' : 'always',
-
-		vite: defineConfig({
-			plugins: [
-				{
-					name: 'watch-templates',
-					configureServer(server) {
-						server.watcher.add('/src/templates');
```
arelra, line 31: Do we not need to pass `/src/templates` for the vite watcher anymore?

SiAdcock: Haha omg you're absolutely right! Thanks for the spot 👍 

It's 2022 and we still have to write our own watch plugins? 😮

arelra (approved):
😎

PR #299, opened 2022-07-27
SPIKE: Migrate interscroller template to Svelte

SiAdcock:
## What does this change?
Migrated the interscroller template to Svelte

The `post` function in `messenger.ts` needed some tweaks to actually be able to work in production. Frontend discards messages that are not strings and ones that do not provide an an id in the format of a UUID.

## How to test
The interscroller makes use of extensive code in frontend (basically all the interscrolling happens there) so unfortunately would be quite difficult to preview using `yarn dev`.

We have setup a test line item accessed with `?adtest=interscroller_test_220727` which you can use on a mobile viewport in a `top-above-nav` slot (first inline).

It should function identically to the legacy template accessable under a different adtest `?adtest=DAP_samung_inter_pg`.

You can verify you're looking at the svelte version by looking for the comment that's inserted at the top:
```
/* "interscroller" updated on 2022-07-27 via https://github.com/guardian/commercial-templates/blob/87a777443/src/templates/ssr/interscroller/index.svelte */
```


src/lib/messenger.ts
```diff
@@ -12,17 +12,45 @@ type StandardMessage<Type = string, Data = unknown> = {
...
+		Math.floor((1 + Math.random()) * 0x10000)
+			.toString(16)
+			.substring(1);
+	return `${_4chars()}${_4chars()}-${_4chars()}-${_4chars()}-${_4chars()}-${_4chars()}${_4chars()}${_4chars()}`;
+};
 
 const post = (arg: Message): void => {
-	window.top?.postMessage({ id: self.name, ...arg }, '/');
+	//  frontend messenger.ts discards messages that are not strings and that do not provide an an id in the format of a UUID
+	window.top?.postMessage(JSON.stringify({ id: generateId(), ...arg }), '*');
```
Jakeii, line 47: pretty sure the `'*'` is safe?

arelra: Yep we do the same thing in Frontend:

https://github.com/guardian/frontend/blob/ebf8b41a14fe7641aa1fa5e49d65b4908cf8c862/static/src/javascripts/projects/commercial/modules/messenger/post-message.ts#L4

SiAdcock (commented):
This looks great to me 👍

Jakeii (approved):

PR #303, opened 2022-08-02
Update the repository to use yarn instead of npm

emma-imber:
## What does this change?

Updated the repository to use yarn instead of npm, to be consistent with the rest of our code.

- Updated to yarn-lock instead of package-lock
- Removed the npmrc file as yarn enforces the engine for us without this
- Update the documentation to use yarn commands instead of npm commands
- Updated the scripts defined in package.json to use yarn
- Added a license to the package.json file (used the same one as commercial-core)

## How to test

Try running yarn or yarn install after cloning the repository and test that the commands work in yarn


README.md
```diff
@@ -17,14 +17,14 @@ found in the [`/legacy` folder](/legacy)
 
 ## Developing Locally
 
-Once you've created a project and installed dependencies with `npm install`,
+Once you've created a project and installed dependencies with `yarn install`,
 start a development server:
 
 ```bash
-npm run dev
+yarn run dev
```
Jakeii, line 24: you can omit the `run` with yarn :)

README.md
```diff
@@ -17,14 +17,14 @@ found in the [`/legacy` folder](/legacy)
 
 ## Developing Locally
 
-Once you've created a project and installed dependencies with `npm install`,
+Once you've created a project and installed dependencies with `yarn install`,
```
Jakeii, line 20: `install` isn't needed with yarn

Jakeii (approved):
yarn ftw!

README.md
emma-imber: Oh fab, thanks!

package.json
```diff
@@ -1,11 +1,12 @@
 {
 	"name": "@guardian/commercial-templates-builder",
 	"version": "0.0.1",
+	"license": "Apache-2.0",
 	"scripts": {
 		"dev": "NODE_ENV=development vite dev --port 7777",
 		"build": "vite build && touch build/.nojekyll",
 		"publish": "npx gh-pages --dist build --dotfiles",
```
arelra, line 8: Think we can also replace `npx` with `yarn` but please verify!

arelra (approved):
Thanks for tidying this up. Yay for standardisation! 🎉

package.json
emma-imber: Just checked and you're right, thanks Ravi!

PR #310, opened 2022-08-26
Add Svelte migration script

SiAdcock:
## What does this change?

Add a script that partially automates the migration of commercial-templates to Svelte. 

This PR represents a first pass at this script, to see if there's enough value in it.

```sh
$ yarn migrate glabs-native-traffic-driver
```

The HTML, SCSS and test.json is fully automated. JS is copied from `legacy` to the Svelte template's directory. However devs are still required to make manual changes to the JS.

https://user-images.githubusercontent.com/5931528/195858867-a3c65b62-6f6b-4dfd-a47f-8613b2366206.mov

## What isn't this doing?

- ❌ The JavaScript that is migrated will not run: it requires devs to make manual changes. These changes are partially documented in [docs/legacy-to-svelte-migration.md](https://github.com/guardian/commercial-templates/blob/main/docs/legacy-to-svelte-migration.md). I started to use the codemod tool [jscodeshift](https://github.com/facebook/jscodeshift) to transform legacy code into modern code but so many changes would be needed, it felt too big a task. The branch [`sa-jscodeshift`](https://github.com/guardian/commercial-templates/compare/svelte-migration-script...sa-jscodeshift) has my attempt.
- ❌ The `app` and `amp` templates are not migrated – only `web` is currently supported
- ❌ JavaScript is not added to the Svelte template. It is created in `index.ts`. This makes it easier to isolate the manual work that a dev needs to complete.

### DIFF
--- a/README.md
+++ b/README.md
@@ -15,8 +15,8 @@ You can see the current templates on [guardian.github.io/commercial-templates](h
 This project is a rewrite and work-in-progress. Legacy templates can still be
 found in the [`/legacy` folder](/legacy)
 
-- [More information on creating Svelte templates](/docs/svelte-template-authoring.md)
-- [More information on migrating legacy templates](/docs/legacy-to-svelte-migration.md)
+- [Creating Svelte templates](/docs/svelte-template-authoring.md)
+- [Migrating legacy templates to Svelte](/docs/legacy-to-svelte-migration.md)
 
 ## Developing Locally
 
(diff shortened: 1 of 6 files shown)


Jakeii:
Awesome! Maybe whether it should be an`ssr` or `csr` template should be an argument to the command?

Jakeii:
@SiAdcock Should we close this? I think migration is a little more nuanced than we thought it might be!

SiAdcock:
Closing as bloody difficult

PR #314, opened 2022-10-17
Script to automatically upload templates to GAM

Jakeii:
## What does this change?

- Modify the svelte build to also output the code to files on disk so they can be read by the upload script.
- Adding an `ad.json` file to the template directories that contain a key `nativeStyleId` for the corresponding native template in GAM
- Writing as script to upload those templates to GAM using the Ad Manager API, this had to be written in python as there's no JS SDK, it seemed the nicest choice of the options Java/Scala/PHP/Ruby/.NET/Python

By default the SDKs use files to pass the OAuth2 credentials in, I modified to use environment variables as they're better suited to CI/CD.

## Next 
In its current form this PR won't do anything.

In a follow-up PR we can setup the github action, keys, and the `ad.json` files for the templates that we want to deploy automatically.

## Why
Copy and pasting into GAM is not ideal, it is slow, error-prone and it's possible for templates to get out of sync with the code in this repository.

## Screenshot
Screenshot 2023-12-13 at 15 35 22


emma-imber (approved):
Looks good to me! Always nice to see a bit of Python 🐍

scripts/upload/README.md
```diff
@@ -0,0 +1,16 @@
+# Upload to GAM
+This script will upload any templates with an `ad.json` containing a `nativeStyleId` property to GAM, replacing the corresponding generated HTML & CSS.
```
SiAdcock, line 2:
```suggestion
This script will upload all commercial templates to GAM, replacing the corresponding generated HTML & CSS.

The script checks for the presence of an ad.json file in the template directories. This file must contain the key `nativeStyleId`, specifying the ID for the corresponding [native style][native-style] in GAM.

[native-style]: https://support.google.com/admanager/answer/13404315?hl=en&ref_topic=7032550&sjid=6297647672569553146-EU
```

PR #317, opened 2022-10-19
Migrate Standard Fabric template to Svelte

Onelittlewolfe:
## What does this change?

Migrating our standard Fabric template to Svelte.

Removing unused animation feature(Too restrictive and better served via the fabric custom).

fabric-animation-fields

Can upload x2 images.

Can have parallax scroll on mobile. 

## How to test

This page is using the new template

Top slot x2 images
Middle parallax scroll for desktop and mobile 
Bottom slot x1 images

https://www.theguardian.com/uk/culture?adtest=svelte_fabric

Need to check that the update doesn't conflict with lineitems using the current Fabric template. It should be ok, as we have updated templates before.

## Accessibility


Jakeii:
Hey Scott, thanks for taking the time to do the conversion! 

I think this template may be better served as a CSR template, so that svelte can take care of the DOM stuff, @mxdvl could you weigh in? I'm not totally sure.

Onelittlewolfe:
> Hey Scott, thanks for taking the time to do the conversion!
> 
> I think this template may be better served as a CSR template, so that svelte can take care of the DOM stuff, @mxdvl could you weigh in? I'm not totally sure.

Hi 

I built the fabric under SSR because it's not dynamic, doesn't pull any api's.

But really aren't all the banners client side rendered though. They might pull data from a server but they are all rendered client side I think.

Jakeii:
> > Hey Scott, thanks for taking the time to do the conversion!
> > I think this template may be better served as a CSR template, so that svelte can take care of the DOM stuff, @mxdvl could you weigh in? I'm not totally sure.
> 
> Hi
> 
> I built the fabric under SSR because it's not dynamic, doesn't pull any api's.
> 
> But really aren't all the banners client side rendered though. They might pull data from a server but they are all rendered client side I think.

Yeah, I'd consider the fact that it's rendering/behaviour changes based on the value of the macros makes it more suitable for client side rendering.

mxdvl:
Hey! I’m no longer on the commercial team, so I’m a little streched to address some of this, but here are some thoughts:

- If you want to benefit most from Svelte, you should handle all DOM manipulation in the Svelte template. This does mean that you would use the CSR approach. SSR templates area only meant for static HTML where the DOM is not manipulated–their only benefit over static HTML is the ability to benefit from Svetle’s scoped styles, sharing components and getting GAM variable intellisense.
- Svelte scopes styles by default and is therefore able to reduce the amount of CSS needed. By importing large shared SASS modules, you lose a lot of the benefits given by the components approach, where styles are colocated with their usage.
- Wherever possible, use components for things that might get reused! A good example could be `<Tracking />` instead of ``
- Helpers such as until-mobile should probably be expressed relying on Source and [named slots](https://svelte.dev/tutorial/named-slots):

```svelte
<script lang="ts">
	import { from } from '@guardian/source-foundations';
</script>

{#if window.matchMedia(from.tablet)}
	<slot name="tablet" />
{:else}
	<slot name="mobile" />
{/if}
```

Jakeii:
Thanks for the insight @mxdvl! 

@Onelittlewolfe I'd be interested in pairing on this, potentially others on comm-dev too, as it's a great case (as max has pointed out) where we can take advantage of various svelte feature.

Onelittlewolfe:
> Thanks for the insight @mxdvl!
> 
> @Onelittlewolfe I'd be interested in pairing on this, potentially others on comm-dev too, as it's a great case (as max has pointed out) where we can take advantage of various svelte feature.

Thanks @mxdvl 

And yes happy to pair on this. For this template, from what Max has pointed out, there's definitely a better way of doing it using svelte features.

Jakeii:
Closing in favour of #339 which used lots of the work layed down here, thanks @Onelittlewolfe!

PR #318, opened 2022-10-19
Fix misalignment of merchandising slots on firefox

emma-imber:
## What does this change?

Adds a margin of zero to make sure that the browser doesn't add a margin around the merch slot content, causing misalignment.

Before:
Screenshot 2022-10-18 at 15 05 24

After:
Screenshot 2022-10-19 at 17 24 34


emma-imber:
> Will this affect templates than merchandising?

Fair point - margin:0 is being automatically applied by chrome on most templates but it wouldn't hurt to make this css change more specific for now, to avoid edge cases

legacy/src/capi-multiple-hosted/web/index.scss
```diff
@@ -28,3 +28,7 @@
     fill: #ffffff;
   }
 }
+
+body {
```
arelra, line 32: Entirely your call but is there scope to create a shared `_merchandising` CSS file from which any shared styling can be imported?

domlander:
> > Will this affect templates than merchandising?
> 
> Fair point - margin:0 is being automatically applied by chrome on most templates but it wouldn't hurt to make this css change more specific for now, to avoid edge cases

Ah OK interesting. Maybe it's a good thing to set margin:0 generally if Chrome is doing that automatically, then deal with edge cases more specifically. Will leave it to your judgement.

domlander (approved):

legacy/src/capi-multiple-hosted/web/index.scss
emma-imber: Yep that's probably a good shout! I'll update accordingly

PR #321, opened 2022-12-06
Fix capi single paidfor logos smaller than 200px stretching

Jakeii:
previously logos were forced to 200px wide, and max height of 60px, causing a stretch with narrower logos.

## Before
image (3)

## After
Screenshot 2022-12-06 at 14 37 53


chrislomaxjones (approved):
👍

ashishpuliyel:
Nice one. I think the US team were asking about this.

PR #328, opened 2023-02-28
Fix fabric margins on Firefox

emma-imber:
## What does this change?

On Firefox, the default browser margin means that fabric ads display with space between the ad label and top of the page (see image), which they shouldn't. This PR sets the margin to zero to ensure that fabric ads display as expected on Firefox.

Screenshot 2023-02-28 at 15 18 36


arelra (approved):
🦊

chrislomaxjones (approved):
Nice spot 👌

PR #329, opened 2023-07-03
Use backticks to fix unexpected token error

emma-imber:
## What does this change?
The capi single paid-for template was broken when the first capi entry had a video associated with it. This is because the videoIcon was undefined, due to speech marks breaking up the variable assignment. Adding backticks to the var declaration ensures the video icon is correctly defined, and that the creative can be rendered.

VideoIcon was defined like this:
Screenshot 2023-07-03 at 17 50 37
With this PR it becomes:
Screenshot 2023-07-03 at 17 51 40

## Images
Creative doesn't load any content:
Screenshot 2023-07-03 at 17 50 22
Creative loads correctly:
Screenshot 2023-07-03 at 17 48 35

### DIFF
--- a/legacy/src/capi-single-paidfor/web/index.html
+++ b/legacy/src/capi-single-paidfor/web/index.html
@@ -53,8 +53,8 @@
   <div class="adverts__body"></div>
 </aside>
 <script>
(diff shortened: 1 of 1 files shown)


chrislomaxjones (approved):
Very nice!

PR #330, opened 2023-07-12
JS tracking field - added at request from Programmatic ops

Onelittlewolfe:
## What does this change?

Adops often have to added JS tracking tags to creatives. Currently Fabrics have this field, but the two templates programmatic ops use( programmatic-mpu and  programmatic-dmpu) do not. This changes brings those templates inline with the Fabrics.

## How to test

NPM run preview 

check the programmatic-mpu  and programmatic-dmpu preview in dev tools. Should be able to the test JS tag.

Picture 29


arelra (approved):
Thanks for keeping the templates in sync!

PR #338, opened 2023-12-21
Convert manual-multiple template to svelte

Jakeii:
## What does this change?

This required a few tweaks to the the `ManualCard` component to make it even more reusable.

Screenshot 2023-12-21 at 17 13 22


src/lib/gam.ts
```diff
@@ -1,9 +1,9 @@
 const CLICK_MACRO = '%%CLICK_URL_UNESC%%';
 const CACHE_BUST = '%%CACHEBUSTER%%';
 
-type GAMVariable<T extends string = string> = `[%${T}%]`;
+type GAMVariable<T extends string = string> = T;
```
Jakeii, line 4: this is necessary so that the template parameters can be used for logic, as typescript complains about stuff like thsi otherwise.
```js
export let Var: GamVariable;

// ts complains because 'true' can never satisfy `[%${T}%]`
if (Var === 'true') {
//
}
```

src/templates/components/ManualCard.svelte
```diff
@@ -1,50 +1,40 @@
...
-	if (regularTitle) {
-		boldTitle += ':';
-	} else {
-		regularTitle = '';
-	}
+	export let isProminent = false;
 </script>
 
 <a
-	class="card split-into-{TotalCardNumber}"
```
Jakeii, line 28: The split into stuff didn't actually work, seems better to filter the events that actually have text in them https://github.com/guardian/commercial-templates/pull/338/files#diff-e5ba83f56a29bdc63ad7241c5b53c92078a77913d1af94896e4a83c974231bf6R69

src/templates/components/ManualCard.svelte
```diff
@@ -121,26 +106,21 @@
...
 	@media (min-width: 740px) {
-		a.card:nth-child(n) {
+		a.card {
+			max-width: 25%;
 			padding: 12px 10px;
 			display: block;
 			margin: 0px;
 		}
 
-		a.split-into-2 {
```
Jakeii, line 131: this is handled by flexbox

src/templates/components/ManualCard.svelte
```diff
@@ -1,50 +1,40 @@
...
+			<img src={image} alt="" />
 		</picture>
 	</div>
+
 	<div class="text">
-		<h2><b>{boldTitle}</b>{regularTitle}</h2>
-		<p>{EventDateTime}</p>
+		<h2><slot name="title" /></h2>
+		<p><slot name="text" /></p>
+		<span class="button" style:--button-colour={buttonColour}>
```
Jakeii, line 30: This can't be an `a` element as we're already inside one

src/templates/csr/events-multiple/index.svelte
```diff
@@ -53,20 +54,44 @@
...
+		buttonText={HeaderButtonText}
+		buttonUrl={HeaderButtonUrl}
+		backgroundColour="#c83877"
+		buttonColour="#ffffff"
+	>
+		<svelte:fragment slot="logo">
+			<LiveLogo />
+		</svelte:fragment>
+		<svelte:fragment slot="description">
+			Our events are a unique opportunity to hear Guardian journalists discuss
```
emma-imber, line 73: The description should be set to the BannerDescription variable as it's something that we manually choose

src/templates/csr/manual-multiple/test.json
```diff
@@ -0,0 +1,30 @@
+{
+	"TrackingId": "",
+	"Tone": "subs-rebrand",
+	"NumberofCards": "3",
```
emma-imber, line 4: Do we still need this variable?

src/templates/csr/manual-multiple/index.svelte
```diff
@@ -0,0 +1,137 @@
+<script lang="ts">
+	import type { GAMVariable } from '$lib/gam';
+	import '$templates/components/fonts/Sans.css';
+	import ManualCard from '$templates/components/ManualCard.svelte';
+	import ManualHeader from '$templates/components/ManualHeader.svelte';
+	import Resizer from '$templates/components/Resizer.svelte';
+	import ToneLogo from '$templates/components/ToneLogo.svelte';
+
+	// export let Title: GAMVariable; // Unused??
```
emma-imber, line 9: We should delete this if it isn't used

src/templates/csr/manual-multiple/test.json
domlander, line 4: `NumberofCards` isn't used in the template. Can it be removed from here?

src/templates/csr/manual-multiple/index.svelte
```diff
@@ -0,0 +1,137 @@
...
+	offers = offers.filter((offer) => offer.title !== '');
+
+	let height: number = -1;
+</script>
+
+<aside bind:clientHeight={height}>
+	<ManualHeader
+		buttonText={ViewAll}
+		buttonUrl={TitleURL}
+		backgroundColour="#005689"
```
emma-imber, line 78: Can we make the header colour also determined by tone? The header colour is just staying fixed as blue even when the tone is changed and the colour should update

src/templates/csr/manual-multiple/index.svelte
```diff
@@ -0,0 +1,137 @@
...
+		buttonColour="#ffe500"
```
emma-imber, line 79: The same comment as above applies to the button colour - this is dependent on tone

src/templates/csr/manual-multiple/index.svelte
domlander, line 9: Can we check with Scott Wolfe whether this is ever used? Otherwise let's remove this

src/templates/components/ToneLogo.svelte
```diff
@@ -0,0 +1,40 @@
...
+	import GenericSvg from './icons/GenericLogo.svelte';
+	import SubscriptionSvg from './icons/SubscriptionsLogo.svelte';
+
+	const toneLogoMapping = {
+		job: JobsSvg,
+		live: LiveSvg,
+		travel: TravelSvg,
+		money: MoneySvg,
+		book: BookSvg,
+		masterclass: MasterclassSvg,
```
emma-imber, line 22: Masterclasses have closed down so they don't need to be listed here

src/templates/csr/events-multiple/index.svelte
```diff
@@ -4,11 +4,12 @@
 	import ManualCard from '$templates/components/ManualCard.svelte';
 	import EventsHeader from '$templates/components/EventsHeader.svelte';
```
domlander, line 5: `EventsHeader` doesn't appear to be used

domlander:
The button on the first card looks a bit awry when `isProminent` is set to true. Could we move it to the bottom-left of the card, so it's consistent with the other cards?

https://github.com/guardian/commercial-templates/assets/9574885/149921d2-22b3-43f2-bdb7-d8d8a8736d8a

src/templates/components/ManualCard.svelte
```diff
@@ -1,50 +1,40 @@
...
+	export let isProminent = false;
 </script>
 
 <a
-	class="card split-into-{TotalCardNumber}"
-	href={EventUrl}
+	class="card"
+	class:is-prominent={isProminent}
+	href={url}
 	style={`--direction: ${direction}`}
```
domlander, line 19: What does this do?

src/templates/components/ManualCard.svelte
```diff
@@ -1,50 +1,40 @@
...
 >
 	<div class="media">
 		<picture>
-			<img src={EventImage} alt="" />
```
domlander, line 34: In the future, it would be good to offer alt text as a GAM variable, for when the image doesn't load

src/templates/csr/manual-multiple/index.svelte
Jakeii: Scott doesn't deal with the merch ads, just the DAP ones like fabric, the variable doesn't appear in the old template so it can't have been used.

src/templates/csr/events-multiple/index.svelte
Jakeii: I'm not sure where this text came from 😅

src/templates/components/ManualCard.svelte
Jakeii: Nothing it seems!

Jakeii:
> The button on the first card looks a bit awry when `isProminent` is set to true. Could we move it to the bottom-left of the card, so it's consistent with the other cards? Or closer to the text above it? It feels a bit like it's floating in space at the moment
> 
> image

I agree it looks odd, I've played around with it and I just can't get it to look good 😂

src/templates/csr/manual-multiple/test.json
emma-imber, line 1: Looking at the current manual multiple templates in ad manager, it seems very common for the variables to include HTML tags.
Screenshot 2024-01-02 at 16 01 29
When I try this locally, the template doesn't recognise these tags.
Screenshot 2024-01-02 at 16 02 15
Could we add support for this? It's fairly straightforward, I think we can add @html in front of the variable, like so: `{@html articleHeadline}`

Jakeii: great shout!

domlander:
I'm not getting the header text when running locally

Screenshot 2024-01-04 at 10 33 20

domlander (approved):

Jakeii:
Tested with about 10 examples that were live at time of writing, and it looks good to go!

PR #339, opened 2023-12-27
Fabric svelte

deedeeh:
## Accessibility


src/templates/ssr/fabric/index.ts
```diff
@@ -0,0 +1 @@
+export {};
```
deedeeh, line 1: I tried to delete this file as we are not using it for Fabric Standard but it gives me an error. Not sure if there is a way to delete it without any compiling errors. I also added `export {}` because it wasn't passing "checking TypeScript" step in CI/CD.

legacy/src/_shared/js/dom.js
```diff
@@ -23,7 +23,6 @@ let img = createElement('img');
 let div = createElement('div');
 
 export {
-    closest,
```
deedeeh, line 26: Dom was adding this for testing but when we paired he told me to remove it because it was causing an error.

src/templates/components/AdvertisementLabel.svelte
```diff
@@ -0,0 +1,51 @@
...
+
+<div class="gs-container label label">
+	<div class="label-text">Advertisement</div>
+</div>
+
+<style lang="scss">
+	@import './colours/palette.scss';
+	$gs-gutter: 20px;
+	$gs-max-columns: 16;
+	$f-sans-serif-text: 'Guardian Text Sans Web', 'Helvetica Neue', Helvetica,
```
Jakeii, line 12: I think we need to import the font file for this

src/templates/components/AdvertisementLabel.svelte
```diff
@@ -0,0 +1,51 @@
+<script lang="ts">
```
Jakeii, line 1: This can go if there's no js/ts

src/templates/components/AdvertisementLabel.svelte
```diff
@@ -0,0 +1,51 @@
...
+</script>
+
+<div class="gs-container label label">
```
Jakeii, line 4: double label class here

src/templates/components/Pixel.svelte
```diff
@@ -0,0 +1,18 @@
+<script lang="ts">
+	import type { GAMVariable } from '$lib/gam';
+
+	export let src: GAMVariable;
```
Jakeii, line 4: as we're in a component here this can be a string type

src/templates/ssr/fabric/index.svelte
```diff
@@ -0,0 +1,154 @@
...
+	export let MobileLayer2BackgroundPosition: GAMVariable;
+	export let MobileLayer3BackgroundImage: GAMVariable;
+	export let MobileLayer3BackgroundPosition: GAMVariable;
+</script>
+
+<AdvertisementLabel />
+<div class="creative creative--fabric">
+	<a
+		id="linkDesktop"
+		class="blink gs-container link hide-until-tablet"
```
Jakeii, line 33: we can merge the classes here that aren't reused for simplicity

src/templates/ssr/fabric/index.svelte
```diff
@@ -0,0 +1,154 @@
...
+	export let MobileLayer1BackgroundImage: GAMVariable;
+	export let MobileLayer1BackgroundPosition: GAMVariable;
+	export let MobileLayer2BackgroundImage: GAMVariable;
+	export let MobileLayer2BackgroundPosition: GAMVariable;
+	export let MobileLayer3BackgroundImage: GAMVariable;
+	export let MobileLayer3BackgroundPosition: GAMVariable;
+</script>
+
+<AdvertisementLabel />
+<div class="creative creative--fabric">
```
Jakeii, line 30: can merge these classes also

src/templates/ssr/fabric/index.svelte
```diff
@@ -0,0 +1,154 @@
...
+		id="linkMobile"
+		class="blink gs-container link mobile-only"
+		href={`${CLICK_MACRO}${ClickthroughUrl}`}
+		target="_blank"
+	>
+		<div class="creative-container">
+			<div class="alt">
+				<div
+					class="layer layer1"
+					style="background-image: url('{MobileLayer1BackgroundImage}');
```
Jakeii, line 67: in svelte you can do `style:background-image={url('{MobileLayer1BackgroundImage}'}` if you prefer, same with background position

src/templates/ssr/fabric/index.svelte
```diff
@@ -0,0 +1,154 @@
...
+			background-image: url('#{dfpVar(MobileLayer1BackgroundImage)}');
+			background-position: dfpVar(MobileLayer1BackgroundPosition);
+		}
+		.layer2 {
+			background-image: url('#{dfpVar(MobileLayer2BackgroundImage)}');
+			background-position: dfpVar(MobileLayer2BackgroundPosition);
+		}
+		.layer3 {
+			background-image: url('#{dfpVar(MobileLayer3BackgroundImage)}');
+			background-position: dfpVar(MobileLayer3BackgroundPosition);
```
Jakeii, line 141: Don't need this css as you've got inline styles above
```suggestion
```

src/lib/messenger/on-scroll.ts
```diff
@@ -0,0 +1,39 @@
+import { postAndListen } from '$lib/messenger';
```
Jakeii, line 1: Although not used in this template good to keep it for the fabric-parallax one!

Jakeii (changes requested):
Nice work @deedeeh! Just a little more tidying up to do

src/templates/ssr/fabric/index.svelte
deedeeh: I don't really understand what do you mean by merge the classes?

src/templates/ssr/fabric/index.svelte
deedeeh: I tried to do that as you suggested but there is an error with url `'url' is not defined`. Could you please expand on the syntax when you have more than one style and also why is there an error?

src/templates/ssr/fabric/index.svelte
deedeeh: This CSS is just for mobile but I wonder if having `dfpVar` makes any difference between the inline styles and this one?

src/templates/ssr/fabric/index.svelte
Jakeii: As in merge the separate classes and styles here into a single class and styles, unless there's an obvious need to reuse one. Looks like `gs-container` doesn't have any styles in this file either?

src/templates/ssr/fabric/index.svelte
Jakeii: Sorry I missed the quotes/backticks, should be ``style:background-image={`url('{MobileLayer1BackgroundImage}')`}``

src/templates/ssr/fabric/index.svelte
Jakeii: the dfpVar wouldn't actually work here anyway, looks like it was a custom sass replacement thing that won't work in svelte https://github.com/guardian/commercial-templates/blob/1bdc306f797c034b6df02bf5c9388bc8a87b754b/legacy/lib/css.js#L33

We don't need it anyway! It's safe to get rid of `dfpVar` and this css

src/templates/ssr/fabric/index.svelte
deedeeh: Ah ok got it. Yes I noticed that now and added it.

deedeeh: `blink`, `gs-container` and `link` are reused in 2 places in this file so not quite sure how to merge them if they are reused. `hide-until-tablet` is this for specific anchor tag.

src/templates/ssr/fabric/index.svelte
deedeeh: I understand your point here and probably in this file we can do that but I guess later on we might have scss files with the reusable properties and I guess creative will have its own class.

src/templates/ssr/fabric/index.svelte
Jakeii: You're right, but all these classes/styles aren't doing very much I think it would be nicer to just a couple more semantic class names like a `link` class for the shared styles and `link-desktop` and `link-mobile` classes for the specific styles.

deedeeh: Refactored with Jake on a call.

src/templates/ssr/fabric/index.svelte
```diff
@@ -0,0 +1,119 @@
...
+		}
+		.layer1 {
+			z-index: 1;
+		}
+		.layer2 {
+			z-index: 2;
+		}
+		.layer3 {
+			z-index: 3;
+		}
```
Jakeii, line 97: Sorry 🙈, I keep seeing optimisations! This css and the classes are actually redundant as [they'll automatically be stacked](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_positioned_layout/Understanding_z-index/Stacking_without_z-index)

src/templates/components/scss/helpers.scss
```diff
@@ -0,0 +1,18 @@
+.gs-container {
```
Jakeii, line 1: I think this file living in `src/styles` or something similar might be better?

src/templates/ssr/fabric/index.svelte
deedeeh: Yes you are totally right!

src/templates/ssr/fabric/index.svelte
```diff
@@ -0,0 +1,110 @@
...
+			background-image: var(--mobile-background-image);
+			background-position: var(--mobile-background-position);
+		}
+	}
+	@media (min-width: 740px) {
+		.layer {
+			background-image: var(--tablet-background-image);
+			background-position: var(--tablet-background-position);
+		}
+	}
```
Jakeii, line 109:
```suggestion
		.layer {
			width: 100%;
			height: 250px;
			position: absolute;
			top: 0;
			left: 0;
			
			background-repeat: no-repeat;
			background-image: var(--mobile-background-image);
			background-position: var(--mobile-background-position);
		}
	}
	@media (min-width: 740px) {
		.layer {
			background-image: var(--tablet-background-image);
			background-position: var(--tablet-background-position);
		}
	}
```

Jakeii (dismissed):

SiAdcock:
Thanks for this @deedeeh 

> This is the Fabric Standard (fixed) which doesn't have animation or parallax. The animation isn't used anymore in this template instead it is used in Fabric Custom and the parallax will have it's own template in a separate PR.

Is the intention that in this PR, we are migrating the existing Fabric template to Svelte, whilst:

- removing the animation functionality as it is no longer used with this template
- preserving the parallax functionality, and moving this into its own Fabric Parallax template in the future ?

This approach would allow us to continue using the same Native Style for this template, as the only breaking change is to remove animation support, which is never used.

Jakeii:
@SiAdcock ah we didn't think that the parallax needed to be done in the same PR, but that makes sense! It needs to be available if we strip it out!

deedeeh:
> Thanks for this @deedeeh
> 
> > This is the Fabric Standard (fixed) which doesn't have animation or parallax. The animation isn't used anymore in this template instead it is used in Fabric Custom and the parallax will have it's own template in a separate PR.
> 
> Just checking my understanding 😄 Is the intention that in this PR, we are migrating the existing Fabric template to Svelte, whilst:
> 
> * removing the animation functionality as it is no longer used with this template
> * preserving the parallax functionality, and moving this into its own Fabric Parallax template in the future ?
> 
> This approach would allow us to continue using the same Native Style for this template, as the only breaking change is to remove animation support, which is never used.

I thought we need to have Parallax in its separate PR from the beginning. I guess I need to understand more about Native Style. Also I am still trying to understand how this will be implemented and how it could be tested locally but I just know that development is very difficult to be tested for Parallax so I would like to pair on it please.

src/templates/ssr/fabric/index.ts
deedeeh: After moving the directory to be a CSR I was able to remove the `ts` file as now all the TypeScript code is in `index.svelte`

src/styles/helpers.scss
```diff
@@ -0,0 +1,28 @@
...
+@function gs-span($n-columns) {
+	@return $n-columns * $gs-column-width + $gs-gutter * ($n-columns - 1);
+}
+
+.gs-container {
+	position: relative;
+	margin: 0 auto;
+
+	@media (min-width: 740px) {
+		padding: 0 calc(50% - #{(gs-span(9) + $gs-gutter * 2) / 2});
```
Jakeii, line 14: I think we simplified this at some point? seems to have been lost?

deedeeh: I made the change and it seems like it is working fine now. Done in 16ee12bd58bab40ad2a19e4e4e64ab5979374f29

src/templates/components/colours/palette.scss
```diff
@@ -0,0 +1,128 @@
+$guardian-brand: #c70000;
```
domlander, line 1: I've created a ticket for us to use this palette, as we currently don't use it

SiAdcock: Some of these colour values could use a [Source makeover](https://theguardian.design/2a1e5182b/p/71fb50-colour/b/399c59) too 💅

src/templates/components/AdvertisementLabel.svelte
```diff
@@ -0,0 +1,28 @@
+<div class="container">
+	<div class="text">Advertisement</div>
+</div>
+
+<style lang="scss">
+	@import './fonts/Sans.css';
+
+	.container {
```
Jakeii, line 8: I think the label won't be aligned with the ad on larger screens without the helpers media queries

domlander: You're right the label is not aligned. I tried it with the media queries but it didn't work. I had to add the padding to replicate the behaviour of the live site.

On a separate note, I notice the layers don't behave correctly with the media queries. I removed them and now I believe the behaviour is aligned to legacy

src/templates/components/colours/palette.scss
domlander: Will add this as a note on the newly created ticket

src/templates/components/AdvertisementLabel.svelte
Jakeii: Without the media queries the ad won't be constrained to certain widths at certain breakpoints like the current fabric is?

domlander: It's a bit confusing because fixed ads behave differently to parallax ads. The background image of the parallax ads is controlled by the commercial code and is outside this template so doesn't respect these media queries.

I've reinstated the media queries, but for parallax ads I've allowed the ad label to go full-width and used padding to place it in the correct position on large screens

Jakeii (approved):
I think it's all good to go, nice work @deedeeh and @domlander!

src/lib/Previews.svelte
```diff
@@ -70,7 +71,7 @@
 
 <section id="example">
 	{#each Object.keys(widths) as width}
-		<div class="size">
+		<div class="size {width === '100%' && 'full-width'}">
```
mxdvl, line 74: In Svelte, this is more semantically achieved by the [`class:` directive](https://svelte.dev/examples/classes)

```svelte
<div class="size" class:full-width={width === 100%}>
```

src/templates/components/Pixel.svelte
```diff
@@ -0,0 +1,16 @@
...
+<style lang="scss">
+	.pixel {
+		position: absolute;
+		bottom: 0;
+		right: 0;
+		width: 1px;
+		height: 1px;
+		opacity: 0;
+	}
+</style>
```
mxdvl, line 16: As Svelte scopes styles automaticall, you can apply the styles directly to `img` without the need for a specific class name.

PR #347, opened 2024-01-04
Convert the CAPI Multiple PaidFor template to Svelte

emma-imber:
## What does this change?
Converts the CAPI Multiple PaidFor template to Svelte. Adds a template type variable to the CapiCard to make it usable in both the single and multiple templates. I've also added an extra breakpoint to the previews as I found it helpful whilst developing.

#### Svelte
svelte360

#### Legacy
Screenshot 2024-01-04 at 14 49 21

#### Svelte
svelte740

#### Legacy
old740

#### Svelte
svelte980

#### Legacy
old980

#### Svelte
svelte1300

#### Legacy
old1300


src/templates/csr/capi-multiple-paidfor/index.svelte
```diff
@@ -0,0 +1,153 @@
...
+
+	let height: number = -1;
+</script>
+
+{#await requestData}
+	<h3>Loading Content for “{SeriesUrl}”</h3>
+{:then response}
+	{#await addHeadlineKicker(cards, response.articles)}
+		<p>Formatting cards...</p>
+	{:then formattedCards}
```
Jakeii, line 78: I think it might be nicer to generate `cards` with a promise? and then await that? 

something like:
```js
const cardOverrides: CapiCardOverride[] = [
//...
];

const getCards = retrieveCapiData(cardOverrides, SeriesUrl).then((response) => addHeadlineKicker(cardOverrides, response.articles));

{#await getCards}
{:then cards}
//
{/await}
```

src/templates/csr/capi-multiple-paidfor/index.svelte
```diff
@@ -0,0 +1,153 @@
...
+			/>
+			<div class="body">
+				<div class="cards-container">
+					{#each formattedCards as single}
+						<CapiCard templateType="multiple" {single} />
+					{/each}
+				</div>
+				<div class="sponsor-container">
+					<Sponsor
+						branding={response.articles[0].branding}
```
Jakeii, line 94: could the response return 0 results?

src/templates/components/CapiCard.svelte
```diff
@@ -1,16 +1,33 @@
 <script lang="ts">
 	import type { Single } from '$lib/types/capi';
 	import Sponsor from './Sponsor.svelte';
+	import AudioIcon from './icons/AudioIcon.svelte';
+	import CameraIcon from './icons/CameraIcon.svelte';
+	import VideoIcon from './icons/VideoIcon.svelte';
 
+	export let templateType: string;
```
Jakeii, line 8: could this be typed to `'multiple' | 'single'`?

src/templates/components/PaidForHeader.svelte
```diff
@@ -14,44 +14,49 @@
 	export let SeriesUrl: GAMVariable;
```
Jakeii, line 14: I think as we're in a component here we shouldn't use `GAMVariable` might lead to unexpected behaviour, like changing the value in the index.svelte but it gets replaced here with the variable content.

src/templates/components/PaidForHeader.svelte
```diff
@@ -14,44 +14,49 @@
...
+						class="popup-link"
+						href={`${CLICK_MACRO}https://theguardian.com/content-funding`}
+						target="_top">Learn more about Guardian Labs content &rarr;</a
+					>
+				</div>
+			{/if}
+		</div>
+
+		<h1
+			class="adverts__title {templateType === 'multiple' && 'multiple-title'}"
```
Jakeii, line 48: In svelte can do stuff like `class:multiple-title={templateType === 'multiple'}` which is a bit nicer.

src/templates/csr/capi-multiple-paidfor/index.svelte
emma-imber: Good point - I'll check the response contains at least one entry before building the templates

src/templates/csr/capi-multiple-paidfor/index.svelte
```diff
@@ -0,0 +1,150 @@
+<script context="module" lang="ts">
+	export const cdn = 'https://i.guim.co.uk/img/media/';
+	export const api =
```
domlander, line 3: Is this being used any more? It looks like this has moved to `capiMultiple.ts`

emma-imber: Great spot, I'll remove it from this file

src/templates/components/PaidForHeader.svelte
```diff
@@ -3,55 +3,60 @@
...
 	import '$templates/components/fonts/SansBold.css';
 
 	import GuardianLabs from './icons/GuardianLabs.svelte';
 
-	export let SeriesUrl: GAMVariable;
-	export let ComponentTitle: GAMVariable;
+	export let SeriesUrl: string;
+	export let ComponentTitle: string;
 	export let edition: Single['branding']['edition'];
+	export let templateType = '';
```
Jakeii, line 16: could have narrower type here too

domlander (approved):

PR #349, opened 2024-01-09
Update SeriesUrl variable name to match format

emma-imber:
## What does this change?
The SeriesUrl variable has been updated to SeriesURL to match the native format. This means the native style should now update automatically via the GitHub Action. It failed on the previous merge because of this variable name mismatch.

Screenshot 2024-01-09 at 15 22 26


arelra (approved):

arelra:
Is there any way we could catch these types of errors in a test?

emma-imber:
> Is there any way we could catch these types of errors in a test?

That's a great idea. I don't think we currently document the native formats at all - maybe there's an argument for storing and updating those automatically like we now do for the templates? Would also allow us to test for any mismatches

PR #350, opened 2024-01-09
Fix Snyk

emma-imber:
## What does this change?
Snyk has been failing on main recently because it hasn't been set up for the new Python scripts for uploading to GAM. This PR adds the parameters we need to handle Python scripts and the Pipfile.

Tested by temporarily running Snyk on push to make sure it now runs successfully:

Screenshot 2024-01-09 at 17 00 39

Screenshot 2024-01-09 at 17 03 47

### DIFF
--- a/.github/workflows/snyk.yml
+++ b/.github/workflows/snyk.yml
@@ -3,8 +3,6 @@ name: Snyk
 
 on:
   push:
-    branches:
-      - main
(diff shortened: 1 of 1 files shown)


arelra (approved):
🐍

Jakeii (approved):

PR #352, opened 2024-01-10
CAPI Multiple PaidFor fixes

emma-imber:
## What does this change?
A number of small fixes for the CAPI Multiple PaidFor template:

- Adds `target="_top" `to the CapiCard link to ensure that articles open up in the full window and not just the ad iframe (see figure 1)
- Adds a margin: 0 to the body so that ads on Firefox display as expected (on Firefox the default margin is not 0, see figure 2)
- Resizes the headers to line up with the lines on the webpage, as the width did not take into account the 40px padding (see figure 3)
- Also set the test.json to an active campaign to help with testing

### Figure 1

The link should open in the main window, not the iframe

Screenshot 2024-01-10 at 14 38 58

### Figure 2

The ad shouldn't have any extra margin around it (Firefox specific issue)

Screenshot 2024-01-10 at 17 44 46

### Figure 3

The right side of the header should line up with the line above it

Screenshot 2024-01-10 at 18 09 01


Jakeii (approved):
Nice 👍

PR #355, opened 2024-01-15
Upgrade Svelte and Sveltekit

Jakeii:
## What does this change?
Upgrade Svelte from v3 to v5 and Sveltekit 1.0.0-next.394 to 2.0.0

We were on an unstable beta version this upgrade to a stable version is long overdue!

The most significant API chages isThe new routing that came with Sveltekit 1.0 https://github.com/sveltejs/kit/discussions/5774, which means

- Different file structure for routes
- no multiple script tags with `context="module"`

Due to the large changes, probably best to wait for any active migrations to finish


src/templates/ssr/interscroller/index.svelte
```diff
@@ -6,14 +6,17 @@
 	<img
 		src="[%TrackingPixel%]"
 		class="creative__pixel creative__pixel--displayNone"
+		alt="pixel"
```
domlander, line 9: I don't think we should have an `alt` tag for these `img`'s. Tracking pixel's aren't displayed to the user so we should probably hide them from screen readers too.

src/templates/components/fonts/Egyptian.css
```diff
@@ -1,6 +1,7 @@
 @font-face {
 	font-family: 'GuardianTextEgyptian';
-	src: url('https://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-Regular.woff2')
+	src:
+		url('https://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-Regular.woff2')
 			format('woff2'),
 		url('https://assets.guim.co.uk/static/frontend/fonts/guardian-textegyptian/noalts-not-hinted/GuardianTextEgyptian-Regular.woff')
```
domlander, line 6: I can't tell what the changes are in these files?

src/templates/components/PaidForHeader.svelte
```diff
@@ -22,10 +18,7 @@
 	<div class="{templateType}-paid-content-and-title">
 		<div class="paid {templateType}-paid-content">
 			<strong>Paid content</strong>
-			<Button
-				aria-controls="popup"
```
domlander, line 26: Why are we losing `aria-controls` and `aria-expanded`?

src/templates/ssr/interscroller/index.svelte
Jakeii: good point!

src/templates/components/fonts/Egyptian.css
Jakeii: the linter has added newlines, now using guardian prettier settings

src/templates/components/PaidForHeader.svelte
Jakeii: redundant as they're already on the `Button` component! https://github.com/guardian/commercial-templates/blob/jlk/upgrade-sveltekit/src/templates/components/Button.svelte#L1

domlander (approved):

src/templates/ssr/interscroller/index.svelte
domlander: I like the `aria-hidden` addition 👍

PR #356, opened 2024-01-15
Set min-height for CapiCard images to 80px

emma-imber:
## What does this change?
Adds a min-height of 80px for images in the Capi Card. Having observed different breakpoints, I can't see any images loading with a height of less than 96px, so this should be a safe value.

We noticed that on the first page load for a CAPI Multiple PaidFor ad on the site, the iframe doesn't load in a big enough size for the ad to show. On subsequent page views it then seems to behave as expected. (Open a lifestyle article in incognito mode to replicate this). The legacy template set min-heights on the images, so we're adding a min-height to images to see if this is the cause of the behaviour or not.

Example of iframe not loading to correct size:
Screenshot 2024-01-15 at 12 51 07

### DIFF
--- a/src/templates/components/CapiCard.svelte
+++ b/src/templates/components/CapiCard.svelte
@@ -49,7 +49,7 @@
 						sizes={source.sizes}
 					/>
 				{/each}
-				<img src={articleImage.backupSrc} alt="" style="min-height: 80px" />
+				<img src={articleImage.backupSrc} alt="" />
 			</picture>
 		{:else}
 			<img src={articleImage.backupSrc} alt="" />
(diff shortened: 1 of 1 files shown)


src/templates/components/CapiCard.svelte
```diff
@@ -49,7 +49,7 @@
 						sizes={source.sizes}
 					/>
 				{/each}
-				<img src={articleImage.backupSrc} alt="" />
+				<img src={articleImage.backupSrc} alt="" style="min-height: 80px" />
```
Jakeii, line 52: does it need to be inline css?

emma-imber: Fair point, it doesn't. Will update!

Jakeii (approved):

PR #357, opened 2024-01-18
Migrate Manual Inline to Svelte

domlander:
## What does this change?

Converts the Manual Inline template to Svelte. 

## Accessibility


src/templates/components/ManualInline.svelte
```diff
@@ -0,0 +1,105 @@
...
+	href={clickMacro(offerUrl)}
+	data-link-name="Offer 1 | {title}"
+	class="card"
+	target="_top"
+>
+	<img src={image} alt={title} />
+	<div class="text">
+		<h2><slot name="title" /></h2>
+		<p class="description"><slot name="text" /></p>
+		<div class="button">
```
Jakeii, line 37: This button might be optional like with `manual-multiple`?

Jakeii (approved):
Just a small thing to check but all looks good, another one bites the dust!

src/templates/components/ManualInline.svelte
domlander: Good point. The button now won't display if the text is blank

PR #360, opened 2024-01-25
Convert Manual Inline to ssr

domlander:
## What does this change?

- Converts the Manual Inline template to SSR.
- The button (div that looks like a button) is now mandatory. This is consistent with the legacy template.

## Why?

- This sidesteps a bug with CSR templates, where if the user add newlines into a user-defined variable in which the html is rendered, the ad will not display.

## How to test

- Follow the steps in the [README to run locally](https://github.com/guardian/commercial-templates/blob/main/README.md#developing-locally)
- Navigate to the Manual Inline template


emma-imber (approved):
Looks good!

Jakeii (approved):

PR #365, opened 2024-02-02
Convert fabric custom template

Jakeii:
## What does this change?
Convert the `fabric-custom` template to svelte.

This template should in theory be CSR, but it has a prop `thirdPartyJSTracking` that can contain unescaped html/js/css which can break the template when rendering (specifically an unescaped `</script>` tag).

There is no way to ensure that the value is escaped as it comes via GAM replacing the value, so I think the only option is to make it SSR so that the prop is not inserted into the script tag, but into the html.

In future we can hopefully get rid of ssr/csr and get the ability to partially prerender templates to solve this problem.


src/templates/components/Pixel.svelte
emma-imber, line 1: On local preview, the pixel is visible underneath the ad: 
Screenshot 2024-02-06 at 13 32 26

Jakeii: That's due to the pixel url in test.json being broken, looks like other templates just have empty strings, so I'll remove them.

domlander (approved):

PR #367, opened 2024-02-14
FIX: Creative opening url in the same iframe

Onelittlewolfe:
## What does this change?

Creative opening url in the same iframe, adding base with a target of _blank is needed.
creative-opening-iniframe

## How to test

Running preview locally clicking banner will open a new window.


Jakeii (approved):
Thanks @Onelittlewolfe, you just need to run `yarn lint:fix` to get it to pass the CI checks then you can merge.

PR #370, opened 2024-02-20
Fix fabric custom script tags

Jakeii:
## What does this change?
script tags in fabric customs weren't running, this was due to a browser security feature that causes script tags inserted via `innerHTML` not to run without any errors or warnings. See https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations

The legacy templates used some sort of hack to get around this, which will still work with svelte.

### DIFF
--- a/src/templates/ssr/fabric-custom/index.ts
+++ b/src/templates/ssr/fabric-custom/index.ts
@@ -23,9 +23,24 @@ const getTag = () => {
 	return Promise.reject('No tag found');
 };
 
+/**
+ * This is a hack to insert the html into the DOM,
+ * using innerHtml would not work if the tag contains a script tag
+ * see https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations
+ *
+ * @param tag the html to insert
+ */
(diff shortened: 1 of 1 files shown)


Onelittlewolfe:
Works for me now with a few different creatives.

src/templates/ssr/fabric-custom/index.ts
```diff
@@ -23,9 +23,24 @@ const getTag = () => {
 	return Promise.reject('No tag found');
 };
 
+/**
+ * This is a hack to insert the html into the DOM,
+ * using innerHtml would not work if the tag contains a script tag
+ * see https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations
```
domlander, line 29: Do the two recommendations: `Element.setHTML()` and `Node.textContent` not work for our use case?

Jakeii: Those would sanitise the html/js, so it wouldn't render just appear as text.

domlander (approved):

PR #372, opened 2024-02-23
Amend logic for hosted advertiser logo

emma-imber:
## What does this change?
The advertiser logo was missing from the hosted template. This is because we were using the [nullish coalescing](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing) operator. The left hand operand should be `null` or `undefined` to return the right hand side. An empty string is falsy but not null, so we were returning an empty string as the logo. I've switched to using a ternary operator, which checks for a falsy value instead.

### Before:

Screenshot 2024-02-23 at 14 52 14

### After:

Screenshot 2024-02-23 at 14 51 14

### DIFF
--- a/src/lib/capiMultiple.ts
+++ b/src/lib/capiMultiple.ts
@@ -29,7 +29,7 @@ function addCapiHostedCardOverrides(
 	overrideCards: CapiCardOverride[],
 	overrideLogo?: string,
 ): { logo: string | null; cards: CapiHostedCard[] } {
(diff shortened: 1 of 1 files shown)


domlander (approved):
I'm not a fan of this rule preventing the logical OR operator in these situations https://typescript-eslint.io/rules/prefer-nullish-coalescing/

PR #374, opened 2024-02-29
Fix column alignment on capi-multiple-hosted template to better align with the DCR grid

i-hardy:
## What does this change?

This amends the spacing and grid layout at the leftcol and wide breakpoints to provide better visual alignment with DCR's grid layout. It also makes some small requested amendments to the styling of card headings, and changes the test data for the template as the Yubico campaign has expired.

## How to test

Viewable with ?adtest=multiple-hosted-alignment

## Wide

Screenshot 2024-02-29 at 13 07 06

## Leftcol

Screenshot 2024-02-29 at 13 08 19


domlander (approved):

HarryFischer:
@i-hardy thanks for this — looks good!

PR #378, opened 2024-04-19
Don't show HTML and CSS by default

emma-imber:
## What does this change?
At the moment, on each template page we show the full HTML and CSS for the template underneath. This is left over from when we had to manually copy the HTML and CSS into the templates in GAM. Now that we don't need to do that, it looks nicer to hide it by default, but it's still easily accessible behind a button.

This change is also a pre-requisite for visual regression testing, as having the HTML and CSS written out is a very common cause of false positives.

### Before:
Screenshot 2024-04-19 at 15 01 15

### After:
Screenshot 2024-04-19 at 14 53 18

Screenshot 2024-04-19 at 14 53 28

### DIFF
--- a/src/lib/Code.svelte
+++ b/src/lib/Code.svelte
@@ -1,6 +1,8 @@
 <script lang="ts">
 	export let html: string;
 	export let css: string;
+	export let showHTML = false;
+	export let showCSS = false;
 </script>
 
 <section id="code">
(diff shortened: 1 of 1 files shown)


src/lib/Code.svelte
```diff
@@ -1,18 +1,38 @@
...
 
 	<div class="html">
 		<h2>HTML</h2>
-		<pre>{html}</pre>
+		<button on:click={() => (showHTML = !showHTML)}>
+			{#if showHTML}
+				Hide HTML
+			{:else}
+				Show HTML
+			{/if}
```
Jakeii, line 18:
```suggestion
			{showHTML ? 'Hide HTML' : 'Show HTML'}
```

src/lib/Code.svelte
```diff
@@ -1,18 +1,38 @@
...
-		<pre>{css}</pre>
+		<button on:click={() => (showCSS = !showCSS)}>
+			{#if showCSS}
+				Hide CSS
+			{:else}
+				Show CSS
+			{/if}
+		</button>
+		{#if showCSS}
+			<pre>{css}</pre>
```
Jakeii, line 34: Only if you feel like it https://svelte.dev/docs/svelte-transition#slide
```suggestion
			<pre transition:slide>{css}</pre>
```
+ above somewhere `import { slide } from 'svelte/transition';` for easy sleek animation!

Jakeii (approved):
Nice!

src/lib/Code.svelte
emma-imber: Ooh nice! I didn't know about this

PR #382, opened 2024-05-22
Fix fabric-custom tracking pixel

Jakeii:
## What does this change?
The `Pixel` component won't work in an SSR component because the [`isValidReplacedVariable`](https://github.com/guardian/commercial-templates/pull/382/files#diff-112bfaccb078632e0383ee5103c3360f7ed59e87ebfc5c543079a469287f88a1L7) used there will cause the `img` element to never render during SSR.

I've moved the pixel code to the `index.ts` and added a guard to the `Pixel` component to prevent it being used in an SSR component in future.

### DIFF
--- a/src/templates/ssr/fabric-custom/index.ts
+++ b/src/templates/ssr/fabric-custom/index.ts
@@ -6,8 +6,8 @@ const DapAssetsFolder: string = '[%DapAssetsFolder%]';
 
 const DapAssetsPath = `${DapAssetsRoot}/${DapAssetsFolder}`;
 const ThirdPartyTag: string = '[%ThirdPartyTag%]';
-const TrackingPixel: string | undefined = '[%TrackingPixel%]';
-const ResearchPixel: string | undefined = '[%ResearchPixel%]';
+const TrackingPixel: string = '[%TrackingPixel%]';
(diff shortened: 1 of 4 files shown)


src/templates/ssr/fabric-custom/index.ts
```diff
@@ -1,10 +1,18 @@
 import { post } from '$lib/messenger';
 
+const CACHE_BUST = '%%CACHEBUSTER%%';
 const DapAssetsRoot = `https://s3-eu-west-1.amazonaws.com/adops-assets/dap-fabrics`;
 const DapAssetsFolder: string = '[%DapAssetsFolder%]';
 
 const DapAssetsPath = `${DapAssetsRoot}/${DapAssetsFolder}`;
 const ThirdPartyTag: string = '[%ThirdPartyTag%]';
+const TrackingPixel: string | undefined = '[%TrackingPixel%]';
+const ResearchPixel: string | undefined = '[%ResearchPixel%]';
```
SiAdcock, line 10: Not a blocker, just for my learning: would these ever be `undefined`?

Jakeii: Ah that was just me testing something! Will remove

emma-imber (approved):
Nice spot!

PR #385, opened 2024-06-18
Use colours from Source palette

emma-imber:
## What does this change?
Where possible, use colours from the Source palette instead of hard coded hex values. I've done this by adding a new file to set CSS variables values using the palette values, and then these CSS var values are imported into each template. Svelte doesn't support CSS in JS, which is how palette works best, but this workaround allows us to still use the palette here without the developer experience being too ugly.

Using the palette means that it's easier to see where the colours come from and why they have certain values, and gives some consistency with other code in the Guardian codebase.

I've checked the templates and everything seems to still look as it should. Some screenshots are below.

### Before:
Screenshot 2024-06-18 at 18 10 45

### After:
Screenshot 2024-06-18 at 18 11 33

### Before:
Screenshot 2024-06-18 at 18 10 59

### After:
Screenshot 2024-06-18 at 18 12 10

### Before:
Screenshot 2024-06-18 at 18 13 09

### After:
Screenshot 2024-06-18 at 18 13 33


src/templates/csr/capi-multiple-paidfor/test.json
```diff
@@ -1,21 +1,21 @@
 {
```
deedeeh, line 1: Is there a reason why we changed the tests to be empty strings?

src/lib/rollup.ts
```diff
@@ -77,6 +78,7 @@ const build = async (
 				},
 			}),
 			typescript({ sourceMap: false }),
+			commonjs(),
```
deedeeh, line 81: I am a bit unclear on why we need `commonjs` from rollup. I know rollup is a bundler but what is considered CommonJS modules in here?

deedeeh (commented):
The screenshots looks identical so that is great! I'm just a bit unclear on couple of things so it would be good to know your thoughts Emma.

src/templates/csr/capi-multiple-paidfor/test.json
emma-imber: They're optional fields so don't need to be specified for the test creative

src/lib/rollup.ts
emma-imber: The source package imports some commonjs so without adding this to rollup the import doesn't get correctly compiled and none of the templates will build. Feel free to play around with it locally but we do need it!

deedeeh: I now understand after you explained that source package imports some commonjs and that's why we obviously need it. I will do that now and let you know if I have more questions.

deedeeh (approved):
Great work Emma 🎉 Thank you for the clarification on the tests and commonjs.

PR #386, opened 2024-06-18
Migrate to use pnpm

emma-imber:
## What does this change?
Migrates from yarn to pnpm in the commercial-templates repository. This aligns us with the `commercial` and `dotcom-rendering` repositories just for some consistency.


package.json
```diff
@@ -50,5 +51,8 @@
 		"vite": "^5.0.12"
 	},
 	"type": "module",
-	"dependencies": {}
+	"dependencies": {
+		"svelte-preprocess": "^6.0.1"
```
Jakeii, line 55: I think this should be a devDependency

package.json
```diff
@@ -3,10 +3,11 @@
 	"version": "0.0.1",
 	"license": "Apache-2.0",
 	"scripts": {
+		"preinstall": "npx only-allow pnpm",
```
Jakeii, line 6: oh what does this do?

Jakeii (approved):

package.json
emma-imber: It gives an error message in the terminal if you try to run yarn instead - handy as I always need the reminder for the first week or so!
Screenshot 2024-06-19 at 10 51 23

PR #390, opened 2024-07-03
Visual regression testing for commercial templates

emma-imber:
## What does this change?
Adds visual regression testing for the Manual Multiple template. Restricting to one template for now to reduce the size of the PR. Works locally and in CI. Compares the local version of the template against the version on main, found at https://guardian.github.io/commercial-templates/

Because we're testing against main there isn't a way to 'approve' a diff other than just merging the PR, but this test should at least alert us if there is a visual diff. If it's one we were planning then that's all good! We can ignore the test failure.

## How to test
Run `pnpm playwright test` locally. Playwright will generate images showing the diff in the `test-results` folder if the test fails.

## Images
I tried changing some text to force a test failure.

Reference image:
Screenshot 2024-07-03 at 15 15 20

Local image:
Screenshot 2024-07-03 at 15 15 25

Diff generated by Playwright:
Screenshot 2024-07-03 at 15 15 31


src/templates/csr/capi-multiple-paidfor/test.json
emma-imber, line 1: I'm updating this now so that I can add testing for this template soon!

playwright/manual-multiple.test.ts
```diff
@@ -0,0 +1,49 @@
...
+			await page.goto(
+				'https://guardian.github.io/commercial-templates/csr/manual-multiple/',
+				{
+					waitUntil: 'networkidle',
+				},
+			);
+
+			const referenceTemplateLocator = page
+				.frameLocator(`[name='width-${breakpoint}']`)
+				.locator('html');
```
emma-imber, line 21: Selecting the html inside the iframe allows us to sidestep issues we can get as a result of padding sometimes being added to the bottom of the iframe, like here:
Screenshot 2024-07-03 at 15 37 33

SiAdcock (approved):
Looking good! 📷 ✨ 👍

PR #391, opened 2024-07-05
Create new test template for Public Good

deedeeh:
## What does this change?

This PR adds a new test for Public Good. A new template named `public-good-test` is created with the new integration as requested by US and Public Good. Check [ticket](https://trello.com/c/MR25vjOO/2949-create-a-new-test-template-for-public-good) for more details and to check the new integration. 

I added the `nativeStyleId` for this native test https://admanager.google.com/59666047#creatives/native/native_style/detail/style_id=844854&tab=ad_settings but this `id` will be changed when we have a PROD Public Good with this new template. 

## How to test

Public Good is served in the `article-end` slot when Reader Revenue epic doesn't show up and it's only served in US so using a VPN is a must to test and add at the end of the article url `?adtest=comm-dev-public-good-test`

## Images

- The new data attributes which doesn't exist in Public Good 

image

- Different height comparing to the Public Good

image

### DIFF
--- a/src/templates/csr/public-good-test/index.svelte
+++ b/src/templates/csr/public-good-test/index.svelte
@@ -18,7 +18,7 @@
 		const consentState = await getUSPData();
 
 		if (!consentState || isCcpaOptedOut(consentState)) {
-			!consentState && console.error('No consent state found');
+			!consentState && console.error('No consent');
 
 			return refresh();
 		}
--- a/src/templates/csr/public-good-test/index.svelte
+++ b/src/templates/csr/public-good-test/index.svelte
@@ -1,0 +1,63 @@
+<script lang="ts">
+	import { getUSPData, isCcpaOptedOut } from '$lib/cmp';
+	import { post } from '$lib/messenger';
+	import { getPageURL } from '$lib/messenger/get-page-url';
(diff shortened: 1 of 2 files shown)


src/templates/csr/public-good-test/index.svelte
```diff
@@ -0,0 +1,63 @@
...
+	const onload = async () => {
+		let url = await getPageURL();
+
+		const consentState = await getUSPData();
+
+		if (!consentState || isCcpaOptedOut(consentState)) {
+			!consentState && console.error('No consent state found');
+
+			return refresh();
+		}
```
arelra, line 24: Very very minor but if `isCcpaOptedOut(consentState)` is `true` we have 'consent state' but consent is not given. So this might be a more accurate log message:

```suggestion
		if (!consentState || isCcpaOptedOut(consentState)) {
			!consentState && console.error('No consent');
			return refresh();
		}
```

arelra (approved):
✨

src/templates/csr/public-good-test/index.svelte
deedeeh: Yes you are right. The message has to be more accurate. Changed in 6bba7c362d5b1c5908f74f23c068358a3e334bb3

PR #392, opened 2024-07-08
Add SSR visual regression tests

emma-imber:
## What does this change?
Adds visual regression tests for Fabric Custom and Manual Single templates. Also refactors the Manual Multiple test to make it quicker - we now have 2 page navigations instead of 8.

## How to test
Run `pnpm playwright test` locally.

Locally I broke the Manual Single template by making the background purple. The test failed, as expected, and the screenshots in the test-results folder show the diff:

Reference:

Screenshot 2024-07-08 at 13 58 01

Local version:

Screenshot 2024-07-08 at 13 58 08

Diff:

Screenshot 2024-07-08 at 13 58 14


playwright/fabric-custom.test.ts
```diff
@@ -0,0 +1,56 @@
...
+				waitUntil: 'networkidle',
+			},
+		);
+
+		for (const breakpoint of widths) {
+			const referenceTemplateLocator = page
+				.frameLocator(`[name='width-${breakpoint}']`)
+				.locator('#creative');
+			// check that the template is present on the page
+			await referenceTemplateLocator.isVisible();
```
arelra, line 23: aiui `isVisible` will evaluate and return immediately rather than waiting. To reduce flakiness its preferable to use `expect(locator).toBeVisible()`

https://playwright.dev/docs/api/class-locator#locator-is-visible

An example:

https://github.com/guardian/dotcom-rendering/blob/db040dce572a80e6d33c71be1b1df006604a6024/dotcom-rendering/playwright/lib/locators.ts#L6

playwright/fabric-custom.test.ts
```diff
@@ -0,0 +1,56 @@
...
+				animations: 'disabled',
+				path: `./playwright/reference-images/Fabric-custom-${breakpoint.replace('%', '')}.png`,
+			});
+		}
+	});
+
+	test('Compare PR templates to reference screenshots', async ({ page }) => {
+		await page.setViewportSize(viewport);
+
+		await page.goto('http://localhost:7777/ssr/fabric-custom', {
```
arelra, line 37: Minor - is it worth creating a shared const for the port number

playwright/manual-single.test.ts
```diff
@@ -0,0 +1,55 @@
+import { expect, test } from '@playwright/test';
+
+const viewport = { width: 1600, height: 1000 };
+
+const widths = ['360', '740', '980', '1300', '100%'];
```
arelra, line 5: Minor - so feel free to ignore. We could get breakpoints from [source-foundations](https://github.com/guardian/source/blob/d84c25d67ed51545994f53986eb4487d35c1aab4/packages/%40guardian/source-foundations/src/breakpoints/breakpoints.ts#L15)

playwright/fabric-custom.test.ts
```diff
@@ -0,0 +1,56 @@
...
+		await page.setViewportSize(viewport);
+
+		await page.goto(
+			'https://guardian.github.io/commercial-templates/ssr/fabric-custom/',
+			{
+				waitUntil: 'networkidle',
+			},
+		);
+
+		for (const breakpoint of widths) {
```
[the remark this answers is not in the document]
emma-imber: The viewport needs to be a minimum width to view some of the template previews hence why we set that at the start. The only relation between that and the widths/breakpoints is that it needs to be bigger than 1300px, the widest preview size.

The breakpoints correspond to the different sizes of template previews we have on one page. So we load the page, and then iterate through the width values because we can use these as selectors for the different iframes.

So the 1300 value is used to select the iframe containing the 1300 sized template preview. So we're iterating through these different iframes:
Screenshot 2024-07-08 at 16 32 35

playwright/fabric-custom.test.ts
emma-imber: Oh I didn't realise that! Good catch, thanks Ravi

playwright/fabric-custom.test.ts
emma-imber: Yeah good shout! I'll do the same for the templates site as well

playwright/manual-single.test.ts
emma-imber: I think because of the 100% value in there not being in source I might leave this for now. But I have extracted the widths into a shared const instead so that they're only defined in one place. I've also updated the variable naming to make it a bit clearer what these sizes actually represent

playwright/fabric-custom.test.ts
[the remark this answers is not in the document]
arelra: I see thanks for explaining!

arelra (approved):
Great work 🎉

PR #395, opened 2024-08-02
Setup CD for legacy templates

Jakeii:
## What does this change?
modify the legacy build script to also copy `ad.json`

modify the deployment script to locate legacy templates with an `ad.json` and upload them to GAM.


arelra (approved):
Great work!

One for a follow up but it might help to add some high level comments to `build.js` and `deploy.py` to add understanding of how they work internally.

emma-imber (approved):

PR #396, opened 2024-08-09
Add PR comment with visual test results

emma-imber:
## What does this change?
Updates the way that visual regression testing works to improve the developer experience. The visual regression test action no longer fails if there are visual changes, to reflect the fact that these may be intentional. Instead, a comment is added to the pull request to indicate if tests are failing, and for which templates.

Screenshot 2024-08-09 at 12 48 09

### DIFF
--- a/playwright/manual-single.test.ts
+++ b/playwright/manual-single.test.ts
@@ -44,7 +44,7 @@ test.describe('Manual Single', () => {
 			// compare screenshot to reference
 			await expect(testTemplateLocator).toHaveScreenshot(
 				`Manual-single-${width.replace('%', '')}.png`,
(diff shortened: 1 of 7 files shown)


playwright/manual-single.test.ts
```diff
@@ -44,7 +44,7 @@ test.describe('Manual Single visual regression testing', () => {
 			// compare screenshot to reference
 			await expect(testTemplateLocator).toHaveScreenshot(
```
arelra, line 45: Very cool! 😎

arelra (approved):
Nice improvement!

Chromatic need to watch out :)

PR #397, opened 2024-08-09
Add more visual regression tests

emma-imber:
## What does this change?
Adds 3 new visual regression tests:
- Fabric
- CAPI Multiple Hosted
- CAPI Multiple Paidfor

Also changes the test.json for both CAPI ads. This is to use an older, more stable campaign for both. This should mean we don't need to keep updating them when they expire, and it also means we can get consistent results for visual testing, instead of noise when we have to update a campaign.


emma-imber:
> ### Visual regression testing results 🔍
> If any tests are failing, please check that any visual changes are intentional before merging your PR.
> 
> Template	Visual test status
> CAPI Multiple Hosted	❌
> CAPI Multiple Paidfor	❌
> Fabric Custom	✅
> Fabric	✅
> Manual Multiple	✅
> Manual Single	✅

The failures for the CAPI ads are expected as we're updating the card content for both in this PR, but they'll be much more stable going forward

dskamiotis (approved):
Tested locally and indeed as expected the CAPI ads are failing.
Assuming once merged the expected images will be updated.
Looks good 👍

PR #401, opened 2024-09-11
Delete all legacy templates that have been migrated to Svelte

emma-imber:
## What does this change?
Deletes any templates in the legacy folder that have been migrated to Svelte, as these are the ones now used by GAM.

Also adds to the visual regression testing, to upload an artifact containing the visual regression test report and the screenshots of any failing tests to help detect the reason for failure. Also updates the default preview iframe height to 650px. This should reduce some test flakiness that happens when the iframe resizing doesn't work as expected.


Jakeii (approved):

playwright/utils.ts
```diff
@@ -2,6 +2,6 @@ const localBaseUrl = 'http://localhost:7777/';
 
 const referenceBaseUrl = 'https://guardian.github.io/commercial-templates/';
 
-const templatePreviewWidths = ['360', '740', '980', '1300', '100%'];
+const templatePreviewWidths = ['740', '980', '1300', '100%'];
```
arelra, line 5: Why did we remove the 360 width from the playwright snapshots?

arelra (approved):
Great! ✨

playwright/utils.ts
emma-imber: Not a permanent change, sorry! Just playing around with the tests atm as they're a little flaky

PR #410, opened 2024-09-27
Migrate fabric video to svelte

dskamiotis:
## What does this change?
Converts the Fabric Video template to Svelte. 

Its important to note, we need to perform setup logic after the component renders. 
- Using the IntersectionObserver, video play logic, depends on the DOM being fully initialised, meant we needed to use the `subscribe` to subscribe to a store, so we can perform actions whenever the value of the store changes.
- ad.json removed from `legacy` fabric video to prevent it being detected by GAM.

## How to test
- Follow the steps in the [README to run locally](https://github.com/guardian/commercial-templates/blob/main/README.md#developing-locally)
- Navigate to the Fabric Video template

We can further check the functionality of the `pause` and the `play` methods by adding console.logs in the "observer" if and navigate the viewport  in Preview so the fabric video is out of view.

We have added the `IntersectionObserver`, which is responsible for detecting when the video element enters the viewport and playing or pausing the video accordingly. Updating the previous functions in the legacy code.

## Images
https://github.com/user-attachments/assets/cff43ea3-bebd-4f47-8dcd-5823b176762c

## Accessibility

-   [x] [The change doesn't use only colour to convey meaning](https://github.com/guardian/accessibility/blob/main/people-and-technology/03-visual.md#use-of-colour)

co-authored: @Jakeii


src/lib/messenger.ts
```diff
@@ -127,5 +128,15 @@ const postAndListen = (arg: Message): Promise<unknown> =>
 		3000,
 	);
 
-export { post, timeout, postAndListen, generateId };
+const onScroll = () =>
+	postAndListen({ type: 'scroll', value: '' }).then(
+		(result) => result as { top: number; bottom: number },
+	);
+
+const onViewport = () =>
```
emma-imber, line 136: I don't see any references to onScroll or onViewport? Could you clarify why these are being added here?

src/templates/csr/fabric-video/index.svelte
```diff
@@ -0,0 +1,219 @@
...
+		style:--desktop-background-position={Layer2BackgroundPosition}
+		style:--mobile-background-image={`url('${MobileLayer2BackgroundImage}')`}
+		style:--mobile-background-position={MobileLayer2BackgroundPosition}
+	/>
+	<div
+		class="layer"
+		style:--desktop-background-image={`url('${Layer3BackgroundImage}')`}
+		style:--desktop-background-position={Layer3BackgroundPosition}
+		style:--mobile-background-image={`url('${MobileLayer3BackgroundImage}')`}
+		style:--mobile-background-position={MobileLayer3BackgroundPosition}
```
emma-imber, line 98: There's lots of duplicate code from the Fabric component - at least 30 lines if not more - could we use that here and add the video as an option inside the component? Feels like it would be easier to maintain if we used the shared component!

src/lib/messenger.ts
```diff
@@ -49,7 +49,8 @@ type StringMessage = StandardMessage<
 	| 'get-page-url'
 	| 'passback-refresh'
 	| 'viewport'
-	| 'scroll',
+	| 'scroll'
+	| 'click',
```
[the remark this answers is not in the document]
dskamiotis: I think this one I discussed with @Jakeii being necessary to track click throughs, as we have a couple or research and tracking props in the Fabric video component.
I assumed it would be as a result of
`const postAndListen = (arg: Message)` where the `Message` type could be a click from the window?
But Im open to be corrected

src/lib/messenger.ts
dskamiotis: I think here also (as above) the functions are designed to send messages to the parent frame and listen for responses, they are not directly called in the provided code, and are exported from this messenger file file so intended to be used elsewhere.

src/templates/csr/fabric-video/index.svelte
dskamiotis: @emma-imber  👍 Refactored to a more generic Fabric Component

src/lib/messenger.ts
dskamiotis: 👍 yes this part of the inital migration but seems latrer the `IntersectionObserver`, which is responsible for detecting when the video element enters the viewport and playing or pausing the video accordingly.

src/templates/csr/fabric-xl/index.svelte
```diff
@@ -28,6 +28,13 @@
 	export let MobileLayer2BackgroundPosition: GAMVariable;
 	export let MobileLayer3BackgroundImage: GAMVariable;
 	export let MobileLayer3BackgroundPosition: GAMVariable;
+	export let VideoURL: GAMVariable;
```
emma-imber, line 31: We don't need to add these variables to the non video templates - we can just give them default values in the shared component

src/templates/components/Fabric.svelte
```diff
@@ -26,33 +34,74 @@
 	export let MobileLayer2BackgroundPosition: string;
 	export let MobileLayer3BackgroundImage: string;
 	export let MobileLayer3BackgroundPosition: string;
+	export let VideoURL: GAMVariable;
```
emma-imber, line 37: Like what's been done with the isXL value - we can give these props a default value so that we don't have to enter them when they're not needed (eg for the Fabric XL and Fabric templates)

src/templates/csr/fabric-video/index.svelte
```diff
@@ -0,0 +1,73 @@
...
+	export let VideoBackupImage: GAMVariable;
+	export let MobileVideoBackupImage: GAMVariable;
+	export let VideoURLMobile: GAMVariable;
+	export let VideoAlignment: GAMVariable;
+
+	export let isXL = false;
+	export let showVideo = true;
+</script>
+
+<Fabric
```
emma-imber, line 42: This file looks great - much neater using the shared component! 🎉

src/templates/components/Fabric.svelte
```diff
@@ -26,33 +34,74 @@
...
-			backgroundRepeat,
-			backgroundPosition,
-		},
+	const video: Writable<HTMLVideoElement | undefined> = writable();
+	let played = false;
+
+	const posterImage = isMobile ? MobileVideoBackupImage : VideoBackupImage;
+	const videoSrc = isMobile ? VideoURLMobile : VideoURL;
+
+	onMount(() => {
```
Jakeii, line 55: I don't think the onMount is needed?

dskamiotis: @Jakeii I think in this case  the onMount function is used to handle video loading, since this Fabric components scope and capabilities have been widened to include the video functionality,  the onMount function is used to conditionally execute code based on the value of showVideo. But there may be a better solution im not aware of?

Jakeii: The `video.subscribe` should take care of it no need for onMount

dskamiotis: @Jakeii I have refacored still taking into account size xl

src/templates/components/Fabric.svelte
```diff
@@ -26,34 +32,78 @@
...
+	export let VideoURL: GAMVariable | undefined = undefined;
+	export let VideoBackupImage: GAMVariable | undefined = undefined;
+	export let MobileVideoBackupImage: GAMVariable | undefined = undefined;
+	export let VideoURLMobile: GAMVariable | undefined = undefined;
+	export let VideoAlignment: GAMVariable | undefined = undefined;
+	export let showVideo = false;
 	export let isXL = false;
 
-	const isMobile = window.matchMedia('(max-width: 739px)').matches;
+	const isMobile = window.innerWidth < 740;
```
Jakeii, line 43: what does this change do? does it need to apply to `isTablet` too?

src/templates/components/Fabric.svelte
```diff
@@ -26,34 +32,78 @@
...
+	if (showVideo) {
+		// eslint-disable-next-line @typescript-eslint/no-unnecessary-condition -- https://github.com/sveltejs/eslint-plugin-svelte/issues/476
+		if (isXL) {
+			post({ type: 'resize', value: { height: 524 } });
+		}
+		video.subscribe((video) => {
+			if (video) {
+				if (!VideoURL || !VideoURLMobile) return;
+
+				video.load(); //
```
Jakeii, line 64: empty comment `//` can be removed and on line below

src/templates/csr/fabric-xl/index.svelte
```diff
@@ -54,5 +56,5 @@
 	{MobileLayer2BackgroundPosition}
 	{MobileLayer3BackgroundImage}
 	{MobileLayer3BackgroundPosition}
-	isXL
+	{isXL}
```
Jakeii, line 59: can be `isXL={true}`

src/templates/csr/fabric-video/index.svelte
```diff
@@ -0,0 +1,73 @@
...
+	export let MobileLayer3BackgroundImage: GAMVariable;
+	export let MobileLayer3BackgroundPosition: GAMVariable;
+
+	export let VideoURL: GAMVariable;
+	export let VideoBackupImage: GAMVariable;
+	export let MobileVideoBackupImage: GAMVariable;
+	export let VideoURLMobile: GAMVariable;
+	export let VideoAlignment: GAMVariable;
+
+	export let isXL = false;
```
Jakeii, line 38: these don't need exported can be inline on the component below

dskamiotis: @Jakeii actually, this one can be removed totally as the Fabric component has a default value of true?

Jakeii: exactly don't need to specify it, if it's the defaults

Jakeii (approved):
Looks good!

PR #415, opened 2024-12-04
Delete gimbap templates

emma-imber:
## What does this change?
These templates haven't been used in GAM in the past year, haven't been edited in years, and DAP are not actively using them, so they can be deleted.

Worst case if we end up needing them, they'll still exist in git history.

Also updates the CAPI paidfor test url - the old one has expired, causing a test failure


deedeeh (approved):

emma-imber:
> I know you said DAP are not actively using them but who knows maybe they might need them in future so do you think an email to them might be a good idea?

I'm going to do some more deletions but can send an email afterwards when they're all done!

PR #416, opened 2024-12-05
Delete remaining unused templates

emma-imber:
## What does this change?
Deletes the following templates, which DAP don't use and that haven't been used in over a year, according to a GAM report:

- blended
- books
- epic
- frame
- google
- jobs
- travel
- truskin-page-skin

Also skips the fabric custom visual regression test because it's been a bit flaky recently. I'll look at fixing it in a 10% time.

🧹

Screenshot 2024-12-05 at 14 26 59


Jakeii (approved):
Nice! 🧹

Jakeii:
Can we delete the travel commercial API too now? 👀 https://github.com/guardian/frontend/blob/main/commercial/conf/routes#L11

emma-imber:
> Can we delete the travel commercial API too now? 👀 https://github.com/guardian/frontend/blob/main/commercial/conf/routes#L11

Potentially! There might still be a test template in GAM using it - I'll add a card to investigate if we can bin it

PR #417, opened 2024-12-09
Migrate fabric-custom-xl with common fabric component

dskamiotis:
## What does this change?
This PR serves as part of the larger body of work to migrate commercial templates from Legacy to Svelte.

Added:
-  New `svelte` ready template 
- Extracted the common HTML structure and styles from the current fabric-custom template and created a FabricCustom component.
Additionally:
- Removed `ThirdPartyTags` (was used for pulling a creative from a tag, which we don't use)

Addition information:
- Introduced an index.ts file alongside index.svelte in the fabric-custom-xl directory to handle server-side logic, ensuring compatibility with Node.js during server-side rendering (SSR).
- The index.svelte file focuses on client-side rendering, making the code more modular and maintainable by separating server-side tasks from client-side presentation.
This approach ensures that server-side logic is executed in the correct environment while keeping the client-side rendering logic clean and focused.

## How to test
Run locally and check the fabric-custom component renders as previously without any UI or behaviour difference. 

Successfully render a template, identical to the legacy version in both UI and functionality.

## Images
Also tested in GAM with `?adtest=custom-xl-migration-test`
Screenshot 2025-01-07 at 13 01 34

| Before| After |
|---------|---------|
| Screenshot 2024-12-11 at 12 43 13| Screenshot 2024-12-11 at 12 49 40 |

https://github.com/user-attachments/assets/3d8a1c91-c5b2-4b4a-b91b-67e4ffc85797

## Accessibility


src/templates/ssr/fabric-custom-xl/test.json
```diff
@@ -0,0 +1,8 @@
+{
+	"DapAssetsFolder": "GooglePixel2",
+	"ThirdPartyTag": "https://tpc.googlesyndication.com/pagead/imgad?id=CICAgKCrtKGjoAEQARgBMgjYagW3ARF6Fw",
+	"TrackingPixel": "",
+	"ResearchPixel": "",
+	"ViewabilityTracker": "<img src=\"\">",
+	"thirdPartyJSTracking": "<SCRIPT TYPE='application/javascript' SRC='https://pixel.adsafeprotected.com/rjss/st/726370/54949606/skeleton.js'></SCRIPT> <NOSCRIPT><IMG SRC='https://pixel.adsafeprotected.com/rfw/st/726370/54949605/skeleton.gif?gdpr=${GDPR}&gdpr_consent=${GDPR_CONSENT_278}&gdpr_pd=${GDPR_PD}' BORDER=0 WIDTH=1 HEIGHT=1 ALT=''></NOSCRIPT>"
+}
```
emma-imber, line 8: Is this just the same test data as the Fabric Custom? I think it would be preferable to use data that's specific to the XL size (maybe ask Scott?). Otherwise this local preview is just exactly the same as the Fabric Custom, and doesn't check that the XL size behaves as we expect. Would also make any future visual regression testing of this template more meaningful

dskamiotis: Great idea, I've updated the test data by updating the `DapAssetsFolder` to `DAP_ad_awards_XL`, Ive also updated the screenshot above in the description.

src/templates/ssr/fabric-custom-xl/index.ts
```diff
@@ -0,0 +1,67 @@
+import { post } from '$lib/messenger';
+
+const CACHE_BUST = '%%CACHEBUSTER%%';
+const DapAssetsRoot = `https://s3-eu-west-1.amazonaws.com/adops-assets/dap-fabrics`;
+const DapAssetsFolder: string = '[%DapAssetsFolder%]';
+
+const DapAssetsPath = `${DapAssetsRoot}/${DapAssetsFolder}`;
+const ThirdPartyTag: string = '[%ThirdPartyTag%]';
```
emma-imber, line 8: You removed ThirdPartyTag for the Fabric Custom - should it be included here or should it be removed from this template too?

dskamiotis: Ah yes good shout, thanks, Ill add to the description as well to keep a note of this going forward 👍 
There seems to be a difference between  `thirdPartyJSTracking` and `ThirdPartyTag`which are unrelated.

legacy/src/fabric-custom-xl/web/index.js
```diff
@@ -6,7 +6,6 @@ const DapAssetsRoot = `https://s3-eu-west-1.amazonaws.com/adops-assets/dap-fabri
 const DapAssetsFolder = '[%DapAssetsFolder%]';
 
 const DapAssetsPath = `${DapAssetsRoot}/${DapAssetsFolder}`;
-const ThirdPartyTag = '[%ThirdPartyTag%]';
```
emma-imber, line 9: Would it be better to just delete the old template in this PR instead of updating it?

src/templates/ssr/fabric-custom-xl/index.ts
```diff
@@ -0,0 +1,67 @@
...
+};
+
+/**
+ * This is a hack to insert the html into the DOM,
+ * using innerHtml would not work if the tag contains a script tag
+ * see https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations
+ *
+ * @param tag the html to insert
+ */
+const insertTag = (tag: string) => {
```
Jakeii, line 41: Some of these functions are copied from `fabric-custom` right? can we move them to a lib and import them to avoid duplication?

dskamiotis: @Jakeii I've placed the shared logic in the `lib` directory (@lib) for consistency rather than create a new folder in the `SSR` directory - which was causing build issues.

src/lib/fabric-custom-shared.ts
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export const getTag = (DapAssetsFolder: string, DapAssetsPath: string) => {
+	if (DapAssetsFolder) {
+		return fetch(`${DapAssetsPath}/index.html`)
+			.then((res) => res.text())
+			.then((html) => replaceAssetLinks(html, DapAssetsPath));
+	}
+	return Promise.reject('No tag found');
+};
```
Jakeii, line 25:
```suggestion
export const addTrackingPixel = (url: string) => {
	const pixel = new Image();
	pixel.src = url + CACHE_BUST;
};

// relative paths in the CSS need to be replaced with the absolute path
export const replaceAssetLinks = (html: string, DapAssetsPath: string) => {
	const re = /url\('\.\/(.*)'\)/g;
	return html.replace(re, `url('${DapAssetsPath}/$1')`);
};

export const getTag = (DapAssetsFolder: string) => {
	if (DapAssetsFolder) {
		return fetch(`${DapAssetsRoot}/${DapAssetsFolder}/index.html`)
			.then((res) => res.text())
			.then((html) => replaceAssetLinks(html, DapAssetsPath));
	}
	return Promise.reject('No tag found');
};
```

src/templates/ssr/fabric-custom-xl/index.ts
```diff
@@ -0,0 +1,47 @@
...
+const ResearchPixel: string = '[%ResearchPixel%]';
+
+/**
+ * This is a hack to insert the html into the DOM,
+ * using innerHtml would not work if the tag contains a script tag
+ * see https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations
+ *
+ * @param tag the html to insert
+ */
+const insertTag = (tag: string) => {
```
Jakeii, line 21: I think this insertTag function can be moved too?

src/lib/fabric-custom-shared.ts
dskamiotis: Thanks @Jakeii, are we sure we want to remove the DapAssetsPath parameter from the `getTag` function? 
This causes an error as we need it on line19 when calling replaceAssetLinks

dskamiotis: @Jakeii, are we sure we want to remove the `DapAssetsPath` parameter from the `getTag` function, as its causing an error. Needed as an argument on line 19 `.then((html) => replaceAssetLinks(html, DapAssetsPath));` and that value is not in scope in this module.

Jakeii: Ah it'll need to be a local variable inside `getTag` then

dskamiotis: @Jakeii, wont need to receive it somehow from the `src/templates/ssr/fabric-custom-xl/index.ts`? as the variable is set as part of thew component template?

Jakeii: `DapAssetsPath` is made from `${DapAssetsRoot}/${DapAssetsFolder}`, `DapAssetsRoot` is in this file, and `DapAssetsFolder` is available in `getTag` as an argument so can get rid of `getDapAssetsPath`

dskamiotis: Thanks @Jakeii didn't spot that! 👍

Jakeii (approved):
We're there!

PR #423, opened 2024-12-30
Update Fabric visual regression tests

emma-imber:
## What does this change?

Adds visual regression tests for the Fabric XL and Fabric video templates. For the Fabric video templates, we remove the autoplay property from the creative when testing, otherwise we would get false positives caused by differing amounts of progress through the video ads.

Also re-enables the Fabric Custom test for most viewport sizes. We turned off this test as it was being flaky. I've looked into it a little and for some reason, it's always the 980px viewport that fails the test, and the other sizes seem ok. I've re-enabled the tests for the other sizes because it's better to have some of the tests working than none at all.


Jakeii (approved):
Nice! Great work figuring out a work around using other sizes 🕵️‍♀️

PR #434, opened 2025-03-21
Override rollup dependencies to minimum version to fix vulnerability

emma-imber:
The `@types/rollup-plugin-css-only` package lists 1 really old version of rollup as a dependency, which has a high severity vulnerability. The package hasn't been updated in years, so there's no update to resolve the out of date dependency, but we do still need it to handle the types of the css rollup plugin. Overriding the dependency to fix the vulnerability isn't ideal, but I think it's fine in this case, especially since it's a dev dependency.

### DIFF
--- a/package.json
+++ b/package.json
@@ -57,5 +57,10 @@
 		"vite": "^5.4.12"
 	},
 	"type": "module",
-	"packageManager": "pnpm@8.15.5"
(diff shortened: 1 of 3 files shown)


i-hardy (approved):
👍

PR #440, opened 2025-04-29
Remove hardcoded version of `corepack` from workflows

cemms1:
## What does this change?

Unsets hardcoded version of corepack in the Github workflows. It shouldn't be needed any more
Missed a couple from https://github.com/guardian/commercial-templates/pull/438


dskamiotis (approved):
this looks good to me 👍

PR #617, opened 2026-04-20
Image template with tracking for adops

Onelittlewolfe:
# Simple Image template

There has been a few situations with DAP and adops where adops needs to use certain tracking tags that are not available on the standard GAM image template, so they use another Creative template within GAM. But that template doesn't accept x2 images.

This pull request introduces a simple native image template that accepts width and height so it can be any size, can accept x1 and x2 images and has extra tracking.

The choice to create as a native template over just updating the Creative template on GAM is that this can be used by both direct and programmatic teams and keeps the code within this repo.

## Testing 
Adtest is up here. Should be able to see a x1 image or x2
https://www.theguardian.com/books?adtest=native_image

https://admanager.google.com/59666047#delivery/line_item/detail/line_item_id=7280754900&order_id=2462673338&li_tab=settings&sort_by=StartDateTime&sort_asc=false

Test native style is here

id: 962313

https://admanager.google.com/59666047#creatives/native/native_style/detail/style_id=962313&tab=ad_settings

### DIFF
--- a/src/routes/templates/image-native/+page.svelte
+++ b/src/routes/templates/image-native/+page.svelte
@@ -8,8 +8,8 @@
 	let {
 		BackgroundImageX1,
 		BackgroundImageX2,
-		adWidth,
-		adHeight,
+		AdWidth,
+		AdHeight,
 		TrackingPixel,
 		ResearchPixel,
 		ViewabilityTracker,
@@ -18,8 +18,8 @@
 
 <div
 	class="creative creative--image"
-	style:--width={`${adWidth}px`}
-	style:--height={`${adHeight}px`}
+	style:--width={`${AdWidth}px`}
+	style:--height={`${AdHeight}px`}
 >
 	<a
 		class="creative__cta"
@@ -29,7 +29,7 @@
 		<picture>
 			<source srcset={`${BackgroundImageX1}, ${BackgroundImageX2} 2x`} />
 			<!-- svelte-ignore a11y-missing-attribute -->
-			<img src={BackgroundImageX1} width={adWidth} height={adHeight} />
+			<img src={BackgroundImageX1} width={AdWidth} height={AdHeight} />
 		</picture>
 	</a>
 	<!-- svelte-ignore a11y-missing-attribute -->
--- a/src/routes/templates/image-native/+page.svelte
+++ b/src/routes/templates/image-native/+page.svelte
@@ -1,0 +1,73 @@
+<script lang="ts">
+	import { building } from '$app/environment';
+	import type { PageData } from './$types';
+	import { CACHE_BUST } from '$lib/gam';
+
+	export let data: PageData;
+
+	let {
+		BackgroundImageX1,
+		BackgroundImageX2,
+		adWidth,
+		adHeight,
+		TrackingPixel,
+		ResearchPixel,
+		ViewabilityTracker,
+	} = data;
+</script>
+
+<div
+	class="creative creative--image"
+	style:--width={adWidth}
+	style:--height={adHeight}
+>
+	<a
+		class="creative__cta"
+		href="%%CLICK_URL_UNESC%%%%DEST_URL_ESC%%"
+		target="_blank"
+	>
+		<picture>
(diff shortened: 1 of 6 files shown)


GHaberis (approved):
First time reviewing a PR in this repo, but I've cross checked the code with existing templates and everything looks good to me 👍

src/routes/templates/image-native/variables.gam.ts
```diff
@@ -0,0 +1,11 @@
+export const gamVariables = {
+	BackgroundImageX1:
+		'https://adops-assets.global.ssl.fastly.net/dap-fabrics/DAP_image_assets_for_testing/mpux1.jpg',
+	BackgroundImageX2:
+		'https://adops-assets.global.ssl.fastly.net/dap-fabrics/DAP_image_assets_for_testing/mpux2.jpg',
+	adWidth: '970',
```
cemms1: Given these are MPU background image examples we could update this to be MPU sized?

```suggestion
	adWidth: '300',
```

src/routes/templates/image-native/+page.svelte
```diff
@@ -0,0 +1,72 @@
...
+		src="{ResearchPixel}{CACHE_BUST}"
+		class="creative__pixel creative__pixel--displayNone"
+		aria-hidden="true"
+	/>
+	<!-- svelte-ignore a11y-missing-attribute -->
+	<img
+		src={ViewabilityTracker}
+		class="creative__pixel creative__pixel--displayNone"
+		aria-hidden="true"
+	/>
```
cemms1, line 52: This is the same strategy as for the interscroller template so should be fine in GAM, but this causes errors in the preview at the moment. Happy to leave as is and we can find a resolution to this in a separate PR!

src/routes/templates/image-native/+page.svelte
```diff
@@ -0,0 +1,72 @@
...
+	import type { PageData } from './$types';
+	import { CACHE_BUST } from '$lib/gam';
+
+	export let data: PageData;
+
+	let {
+		BackgroundImageX1,
+		BackgroundImageX2,
+		adWidth,
+		adHeight,
```
cemms1: Any reason we've lowercased the first character of these variables?

cemms1 (commented):
@GHaberis thanks for reviewing. There's a couple of things I'd like to update with this PR but we can do these as follow ups separately afterwards.

These include:
- adding a missing `ad.json` definition
- adding snapshot testing by adding a Playwright test for this new template

src/routes/templates/image-native/variables.gam.ts
Onelittlewolfe: Updated with MPU sizes

src/routes/templates/image-native/+page.svelte
Onelittlewolfe: No, no good reason for them to lowercased. Updated now.

src/routes/templates/image-native/+page.svelte
Onelittlewolfe: Yes, it seems to work fine in GAM, but would like to know what is causing the error for the preview.

src/routes/templates/image-native/ad.json
```diff
@@ -0,0 +1,6 @@
+{
+	"nativeStyleId": "986360",
+	"creativeTemplateId": "12534062",
+	"testNativeStyleId": "962313",
+	"testCreativeId": "12532595"
```
cemms1: It's not easy from the naming here (we should make a note to improve the dev experience for this!) but this needs to be the ID of a test creative rather than an ID of the native format

I've created a test one for this new template here: [138555949044](https://admanager.google.com/59666047#creatives/creative/detail/line_item_id=7228230439&creative_id=138555949044)


```suggestion
	"testCreativeId": "138555949044"
```

cemms1 (approved):
Once the test creative ID has been updated in the `ad.json` file, this should be good to go! 🚀 

We can investigate the strange preview behaviour with the cache buster URLs in another PR as it's happening for the interscroller template too. Only seems to affect the preview app so not an immediate issue

PR #624, opened 2026-05-07
Bump `@sveltejs/kit`, add `pnpm` override for `postcss`

cemms1:
## What does this change?

- Bumps version of `@sveltejs/kit` to `2.57.1`
- Adds `pnpm override` for `postcss` to ensure it uses at least version `8.5.10`
- Installs `@types/node`


dskamiotis (approved):
👍

PR #638, opened 2026-06-11
Revert "Bump urllib3 from 2.6.3 to 2.7.0 in /scripts/deploy"

deedeeh:
Reverts guardian/commercial-templates#629


dskamiotis (approved):
👍

deedeeh:
It turns out this is not the cause of the native templates Fabric interscrollers not displaying on the page and it's related to DCR. Still investigating which deploy broke that.

PR #644, opened 2026-06-23
Upgrade to Svelte 5 and fix CAPI native templates bug

DanielCliftonGuardian:
### Upgrade from Svelte 4.2.20 to 5.56.3

Using migration script `npx sv migrate svelte-5` this migrates all components and template routes to Svelte 5 runes ($props, $state, $derived, $bindable) and snippet-based slots this was followed with some manual fixes.

### Fix CAPI hosted/native templates bug

In `vite-plugin-extract-assets.ts`, inline <script> bundles now have every < inside JS string / template literals escaped to \x3c via an AST pass. GAM's SafeFrame native renderer re-parses creatives in a tag-aware mode, so any tag like < (e.g. <header, </h1>, <!) in the minified bundle was being interpreted as markup and the script leaked to the page as raw text. Operators (i<e.length) are left untouched.

### Add unit tests for the vite plugins and run in CI
Added unit tests for escapeScriptMarkup (the CAPI escaping logic). Also fixed the pre-existing transformGAMVariables test, which never ran (it didn't set REPLACE_GAM_VARS, so the plugin returned no transform hook), and corrected the test:vite-plugins script glob (vite/*.test.ts matched no files). These tests now run in CI.

### Pin vitest to a Vite-5-compatible version
The test runner was broken: vitest@4 imports vite/module-runner, which only exists in Vite 6+, while we're pinned to Vite 5 (via @sveltejs/vite-plugin-svelte@4). This crashed at startup with ERR_PACKAGE_PATH_NOT_EXPORTED: './module-runner'. Pinned vitest@3.2.4 (supports Vite 5) so the tests run. The full Vite 6 / vite-plugin-svelte 5 / vitest 4 can be done separately and I have created a ticket here https://app.asana.com/1/1210045093164357/project/1213489522854406/task/1216011553322957?focus=true 

`pnpm build:templates` builds all 18 templates successfully, verified the extracted capi-multiple-hosted bundle evaluates as valid JS.

The previously broken CAPI templates are rendering correctly on live preview url https://www.theguardian.com/uk?google_preview=413-C7OHRAYYq7vp0QYwq9ee2QaIAYCAgJC-2YHJNA&iu=59666047&gdfp_req=1&lineItemId=7342828470&creativeId=138563233413

Existing Playwright suite covers the migrated templates.

## Images

image

## Accessibility

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1213863346256885


src/vite/vite-plugin-extract-assets.ts
```diff
@@ -19,6 +19,101 @@ const extractStyles = (dom: JSDOM) => {
...
+ * in its template strings (e.g. `we("<header><h1>...</h1></header>")`).
+ *
+ * Every such `<` lives inside a JS string or template literal, so we replace it
+ * with its `\x3c` escape. `\x3c` === `<` within a literal, so the runtime value
+ * is unchanged while the served source contains no tag-like `<` for SafeFrame to
+ * trip on. Operator `<` in code is left alone: the AST tells us which `<`
+ * characters are inside literals and which are expressions.
+ *
+ * @param {JSDOM} dom
+ */
```
dskamiotis, line 120: Really helpful JSDocs 👍

package.json
```diff
@@ -49,7 +49,7 @@
 		"@sveltejs/adapter-auto": "7.0.1",
 		"@sveltejs/adapter-static": "3.0.10",
 		"@sveltejs/kit": "2.57.1",
-		"@sveltejs/vite-plugin-svelte": "3.1.2",
+		"@sveltejs/vite-plugin-svelte": "^4.0.0",
```
cemms1: Can we pin this dep like we do for others?

src/vite/vite-plugin-extract-assets.ts
```diff
@@ -19,6 +19,101 @@ const extractStyles = (dom: JSDOM) => {
...
+const escapeInlineScripts = (dom: JSDOM) => {
```
cemms1, line 121: I think it might be really useful to include a unit test to explain what this function does in practice

src/vite/vite-plugin-extract-assets.ts
```diff
@@ -19,6 +19,101 @@ const extractStyles = (dom: JSDOM) => {
...
+	const scripts = dom.window.document.querySelectorAll('body script');
+	scripts.forEach((script) => {
+		const original = script.textContent;
+		if (!original.includes('<')) {
+			return;
+		}
+
+		const ast = parseAst(original);
+		const ranges: Array<[number, number]> = [];
+		collectLiteralRanges(ast, ranges);
```
cemms1: Wonder if we can do this without mutating `ranges` within this `collectLiteralRanges` function as it's not clear the way it's written at the moment that `ranges` is mutated as part of this (though the line below `ranges.sort(...)` does hint at that
Otherwise maybe a comment could help?

src/vite/vite-plugin-extract-assets.ts
```diff
@@ -19,6 +19,101 @@ const extractStyles = (dom: JSDOM) => {
...
+ * Recursively collect the `[start, end)` source ranges of every string literal
+ * and template-literal text chunk (`TemplateElement`) in an ESTree AST. These
+ * are the only places a `<` is literal text rather than a JS operator
+ * (`i < e.length`, `1 << 24`), so they are the only places it is safe to escape.
+ *
+ * @param {unknown} node - An AST node, array of nodes, or primitive.
+ * @param {Array<[number, number]>} ranges - Accumulator of `[start, end)` pairs.
+ */
+const collectLiteralRanges = (
+	node: unknown,
```
cemms1: Maybe we could type this more strictly rather than `unknown`? The JSDoc hints that it's going to be an AST node, array of nodes or primitive. Can we be more specific than `unknown`?

src/vite/vite-plugin-extract-assets.ts
```diff
@@ -19,6 +19,101 @@ const extractStyles = (dom: JSDOM) => {
 	return cssContent.trim();
 };
 
+/**
+ * Recursively collect the `[start, end)` source ranges of every string literal
```
cemms1: Do we definitely mean `[start, end)` here? Why do we have two different types of brackets

src/vite/vite-plugin-extract-assets.ts
```diff
@@ -19,6 +19,101 @@ const extractStyles = (dom: JSDOM) => {
...
+		node.forEach((child) => collectLiteralRanges(child, ranges));
+		return;
+	}
+
+	const n = node as {
+		type?: string;
+		start: number;
+		end: number;
+		value?: unknown;
+	};
```
cemms1: Could we use a type guard rather than a type assertion here?

cemms1 (commented):
Really great solution! 👏  I'm happy for the svelte migration changes to go in with the upgrade + fix so long as we add a test or two to demonstrate how the new vite plugin works

package.json
```diff
@@ -66,13 +66,13 @@
...
 		"superstruct": "2.0.2",
-		"svelte": "4.2.20",
+		"svelte": "5.56.3",
 		"svelte-check": "4.4.6",
 		"svelte-preprocess": "6.0.3",
 		"tslib": "2.8.1",
 		"typescript": "5.9.3",
 		"vite": "5.4.21",
-		"vitest": "4.1.2"
+		"vitest": "3.2.4"
```
DanielCliftonGuardian, line 75: I have got the vite-plugin unit tests running again, but had to pin to vitest@3.2.4 because vitest@4 is incompatible with our current Vite version. I have created a separate ticket to bump vite and vitetest https://app.asana.com/1/1210045093164357/project/1213489522854406/task/1216011553322957?focus=true

src/vite/vite-plugin-extract-assets.ts
DanielCliftonGuardian: 👍 I've updated the comment

dskamiotis (commented):
Is this mainly doing the inline-script escaping fix for GAM, plus the test/CI wiring to make sure that path actually gets covered?

src/vite/vite-plugin-extract-assets.ts
```diff
@@ -19,6 +19,112 @@ const extractStyles = (dom: JSDOM) => {
 	return cssContent.trim();
 };
 
+/**
```
dskamiotis, line 22: Nice, this looks like the core GAM fix - is this where the inline script text gets sanitised before extraction?

src/vite/vite-plugin-extract-assets.test.ts
```diff
@@ -0,0 +1,45 @@
...
+	it('preserves the runtime value: un-escaping "\\x3c" returns the original source', () => {
+		const code = `const html = "<header></header>";`;
+		const escaped = escapeScriptMarkup(code);
+
+		// `\x3c` is the character `<`, so the script's behaviour is unchanged...
+		expect(escaped.replaceAll('\\x3c', '<')).toBe(code);
+		// ...but the served source no longer contains any tag-like `<`.
+		expect(escaped).not.toContain('<');
+	});
+});
```
dskamiotis, line 45: Nice test converage 👍

.github/workflows/PR.yml
```diff
@@ -76,3 +76,26 @@ jobs:
...
+        with:
+          node-version-file: '.nvmrc'
+          cache: 'pnpm'
+
+      - name: Install dependencies
+        run: pnpm install --frozen-lockfile
+        shell: bash
+
+      - name: Run vite plugin tests
+        run: pnpm test:vite-plugins
```
cemms1, line 101: Nice! ✨

cemms1 (approved):
👏 well done, this looks good

PR #650, opened 2026-07-07
upgrade vite and vite-svelte-plugin and vitest

oliverabrahams:
## What does this change?

Updates vite, vite-test and @sveltejs/vite-plugin-svelte

Made some fixes to make the template more stable during visual regression. 

Updated the Tracking pixels in image-native to only show if valid. in the Fabric template remove `else if` and replace with `if` as all of these should show if they exist as far as I can tell.  Gina Cox " sometimes it might only be an impression tracker, then other times it might be 2 or 3 different pixels for various measurements supplied" 

@cemms1 suggests we move to chromatic instead of playwright which seems like a great idea to improve the visual regression UI.

Using only the tests in this Repo ATM. If there is a better way to test please let me know.

Everything continues to work as it was. 

Tracking pixels, research pixels, Viewability Tracking can all be shown at the same time on a Fabric Template.

Would be good to get a steer on this.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,4 @@ node_modules
 !.env.example
 vite.config.js.timestamp-*
 vite.config.ts.timestamp-*
+.idea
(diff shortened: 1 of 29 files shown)


DanielCliftonGuardian (approved):

cemms1 (approved):
👏 nice one!

PR #657, opened 2026-07-22
Typescript deploy gam workflow

oliverabrahams:
## What does this change?

Updates the deploy script to use ts rather than python.

Adds the ability to deploy the test templates using the testNativeStyleIds. 

Adds a workflow that can be triggered by adding the the label `deploy-test` which deploys the test templates.

Deployed locally to the test templates and in this PR to the test templates using the `deploy-test` label. 

The idea of this is to begin to improve the ease of testing these templates. This PR combined with individual line items for each test templates under an ad test should allow quite easy and quick testing of dependancy updates. 

Minimal risk


src/routes/templates/public-good-test/+page.svelte
cemms1, line 1: @oliverabrahams shall we remove public good test in a separate PR? Might be better for cleaner commit history/greater understanding

oliverabrahams: Yea that would be better. The only reason i did it here was because it doesn't have a testNativeStyleId. I will move it into a PR that we can merge before this one :)

PR #662, opened 2026-07-27
remove public-good-test from templates as it is not used 

oliverabrahams:
## What does this change?

Removes the `public good test` template. This template was made temporarily as part of a test but is not used in production. We have set it to inactive and this PR removes it from the directory to clean up the repository.

Added here - https://github.com/guardian/commercial-templates/pull/391


deedeeh (commented):
I wonder before we do this is there a `adtest` for Public Good other than this one? I was actually using it today.

oliverabrahams:
> I wonder before we do this is there a `adtest` for Public Good other than this one? I was actually using it today.

I'm not sure how this will affect what you were doing today, can we chat about it. Maybe we can make an adtest line item for the test styles associated with the public-good template?

deedeeh:
> > I wonder before we do this is there a `adtest` for Public Good other than this one? I was actually using it today.
> 
> I'm not sure how this will affect what you were doing today, can we chat about it. Maybe we can make an adtest line item for the test styles associated with the public-good template?

I was asking in general in case I need an adtest for Public Good. We go go ahead with this one and if I need I can create it like you suggested.

deedeeh (approved):

PR #664, opened 2026-07-28
Add Chromatic trial workflow

DanielCliftonGuardian:
## What does this change?

Adds a Chromatic workflow to publish Storybook and run visual regression tests alongside the existing Playwright tests.

Following the [established `dotcom-rendering` pattern](https://github.com/guardian/dotcom-rendering/blob/main/.github/workflows/dcr-chromatic.yml), the workflow checks out the full git history for accurate baselines, uses TurboSnap to test only affected stories, and auto-accepts changes on main. Visual changes are kept non-blocking during the trial, and the process skips cleanly if the CHROMATIC_PROJECT_TOKEN is unconfigured. Finally, it updates the local Chromatic command and documents how to run both Storybook and Chromatic.

See published stories [here](https://www.chromatic.com/build?appId=6a68b779992dc788e0cb14f4) 

---
- To see the specific tasks where the Asana app for GitHub is being used, see below:
  - https://app.asana.com/0/0/1216959658779076

### DIFF
--- a/.github/workflows/chromatic.yml
+++ b/.github/workflows/chromatic.yml
@@ -1,0 +1,71 @@
+name: Chromatic
+
+on:
+  push:
+    branches:
+      - main
+  pull_request:
+    branches-ignore:
+      - 'gh-pages'
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+jobs:
+  chromatic:
+    name: Chromatic
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


oliverabrahams (commented):
This is awesome. When we complete this work we should add it to the storybooks repo https://github.com/guardian/storybooks

oliverabrahams (approved):

PR #665, opened 2026-07-29
Add Storybook stories for more templates

DanielCliftonGuardian:
## What does this change?

This adds Storybook stories for Fabric XL, Manual Single, Manual Multiple and Image Native using their existing GAM test data.

The shared Chromatic viewport configuration has been moved into the Storybook preview config so all stories are tested at the same widths.

Image Native uses an inline transparent tracking pixel because its unreplaced GAM cache buster causes a Vite error in Storybook. This only affects the story and does not change production behaviour.

### DIFF
--- a/.storybook/preview.ts
+++ b/.storybook/preview.ts
@@ -1,9 +1,19 @@
+import { breakpoints } from '@guardian/source/foundations';
 import type { Preview } from '@storybook/sveltekit';
 import '../src/styles/global.scss';
 
 const preview: Preview = {
 	parameters: {
 		layout: 'fullscreen',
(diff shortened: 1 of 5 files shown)


dskamiotis (commented):
We have a couple of tests failing in CI:

`Fabric`  and `image native` is that expected?

dskamiotis (approved):

DanielCliftonGuardian:
> We have a couple of tests failing in CI:
> 
> `Fabric` and `image native` is that expected?

This PR doesn't touch any code that could effect that logic. Just flaky tests which should be improved by us switching to storybook.

PR #666, opened 2026-07-29
Add capi template stories

DanielCliftonGuardian:
## What does this change?

This adds Storybook stories for the CAPI Single Paid For, CAPI Multiple Paid For and CAPI Multiple Hosted templates.

The stories mock the CAPI request modules and use captured responses from the same campaigns used by the existing Playwright tests. This keeps the rendered copy, images and branding consistent while avoiding live API requests during Chromatic builds.

The Storybook TypeScript configuration now supports the explicit module paths required by Storybook mocks and includes Vite asset declarations for the global SCSS import.


dskamiotis (approved):
This looks in order nice work 👍

